PR #5, opened 2022-03-24
7903132: Replace casts with type test pattern

bowbahdoe:
Replaces a handful of casts following type tests with the equivalent pattern syntax.

I cannot create a matching issue, as I do not have an account on the bug tracker.

---------
### Progress

### Issue
 * [CODETOOLS-7903132](https://bugs.openjdk.java.net/browse/CODETOOLS-7903132): Replace casts with type test pattern

### Reviewers
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - no project role)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jextract pull/5/head:pull/5` \
`$ git checkout pull/5`

`$ git checkout pull/5` \
`$ git pull https://git.openjdk.java.net/jextract pull/5/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 5`

`$ git pr show -t 5`

</details>

<a href="https://git.openjdk.java.net/jextract/pull/5.diff">https://git.openjdk.java.net/jextract/pull/5.diff</a>

</details>

### DIFF
--- a/src/main/java/org/openjdk/jextract/clang/Cursor.java
+++ b/src/main/java/org/openjdk/jextract/clang/Cursor.java
@@ -237,10 +237,8 @@ public final class Cursor {
         if (this == other) {
             return true;
         }
-        if (!(other instanceof Cursor otherCursor)) {
-            return false;
-        }
-        return (Index_h.clang_equalCursors(cursor, otherCursor.cursor) != 0);
+        return other instanceof Cursor otherCursor &&
+                (Index_h.clang_equalCursors(cursor, otherCursor.cursor) != 0);
     }
 
     @Override
--- a/src/main/java/org/openjdk/jextract/clang/Cursor.java
+++ b/src/main/java/org/openjdk/jextract/clang/Cursor.java
@@ -237,10 +237,10 @@ public final class Cursor {
         if (this == other) {
             return true;
         }
-        if (!(other instanceof Cursor)) {
+        if (!(other instanceof Cursor otherCursor)) {
             return false;
         }
-        return (Index_h.clang_equalCursors(cursor, ((Cursor)other).cursor) != 0);
+        return (Index_h.clang_equalCursors(cursor, otherCursor.cursor) != 0);
     }
 
     @Override
--- a/src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
+++ b/src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
@@ -56,7 +56,8 @@ abstract class ClassSourceBuilder extends JavaSourceBuilder {
 
     ClassSourceBuilder(JavaSourceBuilder enclosing, Kind kind, String name) {
         this.enclosing = enclosing;
-        this.align = (enclosing instanceof ClassSourceBuilder enclosing) ? enclosing.align : 0;
+        this.align = (enclosing instanceof ClassSourceBuilder classSourceBuilder)
+                ? classSourceBuilder.align : 0;
         this.kind = kind;
         this.desc = ClassDesc.of(enclosing.packageName(), enclosing.uniqueNestedClassName(name));
     }
--- a/src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
+++ b/src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
@@ -56,8 +56,7 @@ abstract class ClassSourceBuilder extends JavaSourceBuilder {
 
     ClassSourceBuilder(JavaSourceBuilder enclosing, Kind kind, String name) {
         this.enclosing = enclosing;
-        this.align = (enclosing instanceof ClassSourceBuilder) ?
-            ((ClassSourceBuilder) enclosing).align : 0;
+        this.align = (enclosing instanceof ClassSourceBuilder enclosing) ? enclosing.align : 0;
         this.kind = kind;
         this.desc = ClassDesc.of(enclosing.packageName(), enclosing.uniqueNestedClassName(name));
     }
@@ -222,8 +221,8 @@ abstract class ClassSourceBuilder extends JavaSourceBuilder {
 
     ToplevelBuilder toplevel() {
         JavaSourceBuilder encl = enclosing;
-        while (encl instanceof ClassSourceBuilder) {
-            encl = ((ClassSourceBuilder) encl).enclosing;
+        while (encl instanceof ClassSourceBuilder classSourceBuilder) {
+            encl = classSourceBuilder.enclosing;
         }
         return (ToplevelBuilder)encl;
     }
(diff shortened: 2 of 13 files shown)


mcimadamore:
Are you in a position to file a JBS issue for this? We're happy to sponsor - the problem here is that the subject of the PR has to be in the form:

NNNNNNN: <JBS issue subject>

For it to go through. If you are having issues, please let us know and we'll try to find an alternate arrangement.

bowbahdoe:
My understanding is that I can't file a JBS issue until I get an account, which only happens after having made two contributions and petitioning for becoming an author.

JornVernee:
I've filed an issue for this PR here: https://bugs.openjdk.java.net/browse/CODETOOLS-7903132 Please update the PR title to include the issue number.

src/main/java/org/openjdk/jextract/clang/Cursor.java
```diff
@@ -237,10 +237,10 @@ public boolean equals(Object other) {
         if (this == other) {
             return true;
         }
-        if (!(other instanceof Cursor)) {
+        if (!(other instanceof Cursor otherCursor)) {
             return false;
         }
-        return (Index_h.clang_equalCursors(cursor, ((Cursor)other).cursor) != 0);
+        return (Index_h.clang_equalCursors(cursor, otherCursor.cursor) != 0);
```
JornVernee, line 243: I think in these cases we can do better by merging the `instanceof` check above this in the condition. i.e.
```suggestion
        return other instanceof Cursor otherCursor
            && (Index_h.clang_equalCursors(cursor, otherCursor.cursor) != 0);
```

And similar for other places that are computing a boolean value, such as all the equals methods.

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
```diff
@@ -56,8 +56,7 @@
 
     ClassSourceBuilder(JavaSourceBuilder enclosing, Kind kind, String name) {
         this.enclosing = enclosing;
-        this.align = (enclosing instanceof ClassSourceBuilder) ?
-            ((ClassSourceBuilder) enclosing).align : 0;
+        this.align = (enclosing instanceof ClassSourceBuilder enclosing) ? enclosing.align : 0;
```
JornVernee, line 59: This is resulting in a compilation error:
```
jextract\src\main\java\org\openjdk\jextract\impl\ClassSourceBuilder.java:59: error: variable enclosing is already defined in constructor ClassSourceBuilder(JavaSourceBuilder,Kind,String)
        this.align = (enclosing instanceof ClassSourceBuilder enclosing) ? enclosing.align : 0;
```
Please build and test the patch locally on your machine as well, if you haven't done so already. (This should be done before suggesting changes).

bowbahdoe: So yes, 100% on me. But in my defense I am getting a particularly obtuse error before compilation and there isn't exactly someone to ask without pretense.

```
$ /usr/bin/clang --version
Apple clang version 13.1.6 (clang-1316.0.21.2)
Target: arm64-apple-darwin21.3.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

$ sh ./gradlew -Pjdk18_home=/Users/emccue/Library/Java/JavaVirtualMachines/openjdk-18/Contents/Home  clean verify

FAILURE: Build failed with an exception.

* Where:
Build file '/Users/emccue/Development/jextract/build.gradle' line: 13

* What went wrong:
A problem occurred evaluating root project 'jextract'.
> Could not get unknown property 'libclang_home' for root project 'jextract' of type org.gradle.api.Project.

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 297ms

$ sh ./gradlew -Pjdk18_home=/Users/emccue/Library/Java/JavaVirtualMachines/openjdk-18/Contents/Home -Plibclang_home=/usr/bin/clang clean verify

FAILURE: Build failed with an exception.

* Where:
Build file '/Users/emccue/Development/jextract/build.gradle' line: 14

* What went wrong:
A problem occurred evaluating root project 'jextract'.
> Cannot invoke method getAt() on null object

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 287ms

$ sh ./gradlew -Pjdk18_home=/Users/emccue/Library/Java/JavaVirtualMachines/openjdk-18/Contents/Home -Plibclang_home=/usr/bin/ clean verify

FAILURE: Build failed with an exception.

* Where:
Build file '/Users/emccue/Development/jextract/build.gradle' line: 14

* What went wrong:
A problem occurred evaluating root project 'jextract'.
> Cannot invoke method getAt() on null object

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 290ms
```

src/main/java/org/openjdk/jextract/clang/Cursor.java
bowbahdoe: Strictly speaking, the whole method body could be merged into one expression. Its the realm of infinite bikesheds though, and I did not / do not know if there is a set of implicit style choices there.

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
bowbahdoe: (I was counting on some CI/CD to be triggered on PR, which i suppose either isn't the case or isn't the case unless i get past the issue number check)

JornVernee: Questions can be directed at the mailing list (either jextract-dev@openjdk.java.net or panama-dev@openjdk.java.net). You can send an email there without making an account as well (it will go through a moderation queue first).

I agree the error message is not great. In this case the issue seems to be this line in build.gradle:

```
def clang_version = new File("${libclang_home}/lib/clang/").list()[0]
```

i.e. the `libclang_home` directory does not have the expected directory structure. It is looking for a directory like `<libclang_home>/lib/clang/<VERSION>`. So please make sure that exists.

src/main/java/org/openjdk/jextract/clang/Cursor.java
JornVernee: Well, we're on the topic of style changes already. This is new territory, so I think this is where we decide this :)

Merging the `instanceof` into the expression avoids having the pattern variable be carried over to outside of the `if` block, which I think is better.

The `this == other` check above that doesn't have the same problem.

bowbahdoe: I don't have any particularly strong opinions there, so I made that as the change, leaving the `this == other` checks alone

mcimadamore (approved):
Looks good

JornVernee (approved):

PR #9, opened 2022-03-28
7903136: Don't require jdk18_home to be set in jextract gradle build

JornVernee:
Drop the requirement to set `jdk18_home` in the gradle build, and rely instead on `JAVA_HOME` being set to JDK 18.

I've also updated the build instructions. The `-Pjtreg_home` flag seems to have been dropped by accident from the test command. I've re-added it. Also renamed `libclang_dir` -> `llvm_dir` which was missed in an earlier change.

---------
### Progress

### Issue
 * [CODETOOLS-7903136](https://bugs.openjdk.java.net/browse/CODETOOLS-7903136): Don't require jdk18_home to be set in jextract gradle build

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jextract pull/9/head:pull/9` \
`$ git checkout pull/9`

`$ git checkout pull/9` \
`$ git pull https://git.openjdk.java.net/jextract pull/9/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 9`

`$ git pr show -t 9`

</details>

<a href="https://git.openjdk.java.net/jextract/pull/9.diff">https://git.openjdk.java.net/jextract/pull/9.diff</a>

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 `jextract` depends on the [C libclang API](https://clang.llvm.org/doxygen/group__CINDEX.html). To build the jextract sources, the easiest option is to download LLVM binaries for your platform, which can be found [here](https://releases.llvm.org/download.html) (a version >= 9 is required). Both the `jextract` tool and the bindings it generates depend heavily on the [Foreign Function & Memory API](https://openjdk.java.net/jeps/419), so a suitable [jdk 18 distribution](https://jdk.java.net/18/) is also required.
 
-The jdk used to build jextract is specified with `-Pjdk=/path/to/jdk`. If that option is omitted, a build JDK will be downloaded automatically.
+The JDK used to build jextract can be specified with `-Pjdk=/path/to/jdk`. If that option is omitted, a build JDK will be downloaded automatically.
 
 `jextract` can be built using `gradle`, as follows (on Windows, `gradlew.bat` should be used instead):
 
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 `jextract` depends on the [C libclang API](https://clang.llvm.org/doxygen/group__CINDEX.html). To build the jextract sources, the easiest option is to download LLVM binaries for your platform, which can be found [here](https://releases.llvm.org/download.html) (a version >= 9 is required). Both the `jextract` tool and the bindings it generates depend heavily on the [Foreign Function & Memory API](https://openjdk.java.net/jeps/419), so a suitable [jdk 18 distribution](https://jdk.java.net/18/) is also required.
 
-By default, the build will use the JDK used to run gradle. An alternative JDK can be specified with `-Pjdk=/path/to/jdk`.
+The jdk used to build jextract is specified with `-Pjdk=/path/to/jdk`. If that option is omitted, a build JDK will be downloaded automatically.
 
 `jextract` can be built using `gradle`, as follows (on Windows, `gradlew.bat` should be used instead):
 
--- a/README.md
+++ b/README.md
@@ -6,14 +6,14 @@
 
 `jextract` depends on the [C libclang API](https://clang.llvm.org/doxygen/group__CINDEX.html). To build the jextract sources, the easiest option is to download LLVM binaries for your platform, which can be found [here](https://releases.llvm.org/download.html) (a version >= 9 is required). Both the `jextract` tool and the bindings it generates depend heavily on the [Foreign Function & Memory API](https://openjdk.java.net/jeps/419), so a suitable [jdk 18 distribution](https://jdk.java.net/18/) is also required.
 
-`jextract` can be built using `gradle`, as follows (on Windows, `gradlew.bat` should be used instead):
-
 Before starting, please make sure `JAVA_HOME` points and jdk 18:
 
 ```sh
 $ echo $JAVA_HOME
 /path/to/jdk18
 ```
+
+`jextract` can be built using `gradle`, as follows (on Windows, `gradlew.bat` should be used instead):
 
 ```sh
 $ sh ./gradlew -Pllvm_home=<libclang_dir> clean verify
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Expected a header file
 The repository also contains a comprehensive set of tests, written using the [jtreg](https://openjdk.java.net/jtreg/) test framework, which can be run as follows (again, on Windows, `gradlew.bat` should be used instead):
 
 ```sh
-$ sh ./gradlew -Pllvm_home=<libclang_dir> jtreg
+$ sh ./gradlew -Pllvm_home=<libclang_dir> -Pjtreg_home=<jtreg_dir> jtreg
 ```
 
 ### Using jextract
--- a/README.md
+++ b/README.md
@@ -8,8 +8,15 @@
 
 `jextract` can be built using `gradle`, as follows (on Windows, `gradlew.bat` should be used instead):
 
+Before starting, please make sure `JAVA_HOME` points and jdk 18:
+
 ```sh
-$ sh ./gradlew -Pjdk18_home=<jdk18_home_dir> -Pllvm_home=<libclang_dir> clean verify
+$ echo $JAVA_HOME
+/path/to/jdk18
+```
+
+```sh
+$ sh ./gradlew -Pllvm_home=<libclang_dir> clean verify
 ```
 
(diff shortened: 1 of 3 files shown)


JornVernee:
> But maybe there are some advantages - e.g.. maybe in terms of friendliness w.r.t. IDE support?
That too. I looked at it a bit yesterday, and it doesn't seem like IntelliJ supports specifying flags to it's Gradle build.

JAVA_HOME can also be set on shell startup, which would avoid having to set anything at all. Or, at least it would only have to be set once per shell session.

These are quality of life improvements I think.

Being able to manually specify the JDK is reasonable though. I can keep a `-Pjdk` flag, and adjust the version check to check for 18+ instead of exactly 18 (not sure how easy that is for EA versions though, maybe the check should just be dropped).

mcimadamore:
> These are quality of life improvements I think.

I don't object to wanting to improve the quality of the build process experience. At the same time I'm worried about doing something (like this PR does) which might work well in some cases, and not at all in others, at which point you need to have _two_ ways to do the same thing, which leads to a more complex build.

My feeling is that a more principled approach, if we really aim for quality of life, would be to keep the parameters, but add a section where we infer which artifacts (JDK, libclang) we need depending on the arch/OS, and then fetch them automatically (but users can override if they so wish).

That said, I have no super strong opinions on this. I suppose we could also try this out and see how it works out in practice, and correct if/when needed.

sundararajana:
> Being able to manually specify the JDK is reasonable though. I can keep a -Pjdk flag, and adjust the version check to >  check for 18+ instead of exactly 18 (not sure how easy that is for EA versions though, maybe the check should just be > dropped).

But the assumption is that jextract "18" will build and work with jdk18+ - which will not true after foreign-preview branch hits mainline. I think it is better to call "jdk18", "jdk19" explicitly till the point foreign-abi API becomes part of specification in some future jdk. At that point, we can have jdk<N>+ check.

bric3:
As an alternative to property or JAVA_HOME, did you consider using [Gradle java toolchains](https://docs.gradle.org/current/userguide/toolchains.html) ?

mcimadamore:
> As an alternative to property or JAVA_HOME, did you consider using [Gradle java toolchains](https://docs.gradle.org/current/userguide/toolchains.html) ?

This was proposed here:
https://mail.openjdk.java.net/pipermail/panama-dev/2022-March/016665.html

While for regular projects toolchains is great - I'm not sure it seems to limiting here. If I specify `JavaLanguageVersion.of(19)` I get an error:
```
> Unable to download toolchain matching these requirements: {languageVersion=19, vendor=any, implementation=vendor-specific}
```
Which is understandable, given 19 doesn't really exists yet. I think what we'd like would be some kind of toolchain-like solution, but which is backed by https://jdk.java.net/

bric3:
That's indeed an issue but the toolchain allows to specify where to find the JDKs via properties

```properties
org.gradle.java.installations.fromEnv=JDK18,JDK19
```
```properties
org.gradle.java.installations.paths=/custom/path/jdk18,/custom/path/jdk19
```

The benefit is that there's no collusion with `JAVA_HOME` which may be used by something else, in particular Gradle.

----

From what I understand Gradle team is working to provide a better toolchain selection, but it is not there yet.

mcimadamore:
I've tried `org.gradle.java.installations.paths` and that indeed helps finding a non-GA version of the JDK that is locally installed.

Assuming we can make everything work (e.g. moditect plugin does not honor toolchain AFAIK), and assuming that ides work better with toolchains, that works for me. I would disable auto-provisioning out of the box though (users can re-enable locally).

> That's indeed an issue but the toolchain allows to specify where to find the JDKs via properties
> 
> ```ini
> org.gradle.java.installations.fromEnv=JDK18,JDK19
> ```
> 
> ```ini
> org.gradle.java.installations.paths=/custom/path/jdk18,/custom/path/jdk19
> ```
> 
> The benefit is that there's no collusion with `JAVA_HOME` which may be used by something else, in particular Gradle.
> 
> From what I understand Gradle team is working to provide a better toolchain selection, but it is not there yet.

bric3:
> I would disable auto-provisioning out of the box though

That makes sense.

> Amssuming we can make everything work (e.g. moditect plugin does not honor toolchain AFAIK)

I've never used moditect, and I haven't looked up alternatives either, but it might be possible such tools have options like `jvm`, `jdk`, `jdkHome`.

mcimadamore:
> and assuming that ides work better with toolchains

That does not seem to be the case, unfortunately. That is, IDE needs the "installation path" property to be set - but, as for the current property, IntelliJ does not allow to easily add gradle properties. So, from an IDE usability perspective, using toolchains w/o auto-download is the same as what we currently have.

mcimadamore:
After thinking some more, and having tried different things, it seems to me that, while alternatives are possible, nothing seems quite a slam dunk. The best and most consistent way to provide better user experience would be to add our own download logic of binaries in https://jdk.java.net - failing that, all solutions seem to suffer from some pros and cons - which doesn't make them a a big enough improvement over the status quo, IMHO.

bric3:
Fair. I believe tool-chains support will improve in the future. IntelliJ for example don't explicitly support tool-chain but it will follow what is declared in Gradle, so this might be encoded as Gradle task which is indeed inefficient in some cases.
Maybe this can be revisited when tool-chains tooling is more solid and more compelling.

JornVernee:
I've updated the PR and added an automatic download step that downloads JDK 18 from jdk.java.net when `-Pjdk` is not specified.

README.md
```diff
@@ -6,10 +6,12 @@
 
 `jextract` depends on the [C libclang API](https://clang.llvm.org/doxygen/group__CINDEX.html). To build the jextract sources, the easiest option is to download LLVM binaries for your platform, which can be found [here](https://releases.llvm.org/download.html) (a version >= 9 is required). Both the `jextract` tool and the bindings it generates depend heavily on the [Foreign Function & Memory API](https://openjdk.java.net/jeps/419), so a suitable [jdk 18 distribution](https://jdk.java.net/18/) is also required.
 
+The JDK used to build jextract can be specified with `-Pjdk=/path/to/jdk`. If that option is omitted, a build JDK will be downloaded automatically.
+
```
sundararajana, line 10: Should we say JDK used to build jextract should be jdk 18 ?

build.gradle
```diff
@@ -1,23 +1,102 @@
...
+    }
+
+    return ext.depsDir + "/jdk-18";
+}
+
+String jdkHome;
+if (project.hasProperty("jdk")) {
+    println("using JDK from 'jdk' project property");
+    jdkHome = project.property("jdk");
+    checkJDKRelease(jdkHome, minimalJDKVersion);
```
sundararajana, line 93: we check minimum JDK. So.. jdk 19 will be accepted even after foreign-preview hits jdk 19. But that won't work with this version of jdk. I think it is better to accept only jdk 18 - as this jextract code/test/samples work only with jdk 18.

README.md
JornVernee: The jdk 18 requirement is listed in the paragraph before.

JornVernee:
After some consideration and offline discussion. Implementing & maintaining our own JDK download code doesn't seem like it gives enough benefit for the cost.

For now we will stick with users manually having to supply a JDK 18 home. A local `gradle.properties` file could also be used to avoid having to pass this flag every time when invoking gradle. As well as using gradle with IntelliJ, which doesn't support passing command line flags to it's gradle builder (AFAICT).

Maybe once the toolchain feature supports jdk.java.net JDKs out-of-the-box, we could reconsider.

build.gradle
```diff
@@ -1,23 +1,102 @@
...
+def static checkJDKRelease(String jdkHome, Runtime.Version expectedVersion) {
+    Runtime.Version foundVersion = getJDKVersion(jdkHome);
+    if (foundVersion.compareToIgnoreOptional(expectedVersion) < 0) {
+        throw new IllegalArgumentException("Incorrect build JDK version. " +
+            "Expected at least '${expectedVersion}', but found '${foundVersion}'");
+    }
+}
+
+def String downloadJDK() {
+    String jdkUrl = "https://download.java.net/java/GA/jdk18/43f95e8614114aeaa8e8a5fcf20a682d/36/GPL/openjdk-18_";
```
bric3, line 59: You should put this in a property so gradle won't have to recompile/reconfigure the script if this value changes.

PR #11, opened 2022-03-31
7903139: avoid jpackage in build.gradle

sundararajana:
Using jlink with launcher option instead of jpackage. This ensures uniform layout of app dir ($app/bin/jextract script) on all platforms. Also this avoids the use of third-party jpackage grade plugin.

---------
### Progress

### Issue
 * [CODETOOLS-7903139](https://bugs.openjdk.java.net/browse/CODETOOLS-7903139): avoid jpackage in build.gradle

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jextract pull/11/head:pull/11` \
`$ git checkout pull/11`

`$ git checkout pull/11` \
`$ git pull https://git.openjdk.java.net/jextract pull/11/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 11`

`$ git pr show -t 11`

</details>

<a href="https://git.openjdk.java.net/jextract/pull/11.diff">https://git.openjdk.java.net/jextract/pull/11.diff</a>

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,22 +26,8 @@ $ sh ./gradlew -Pjdk18_home=<jdk18_home_dir> -Pllvm_home=<libclang_dir> clean ve
 > 
 > </details>
 
-After building, there should be a new `jextract` folder under `build` (the contents and the name of this folder might vary slightly depending on the platform, e.g. on macOs the folder is `jextract.app`) :
-
-```
-build/jextract
-├── bin
-└── lib
-    ├── app
-    └── runtime
-        ├── bin
-        ├── conf
-        ├── include
-        ├── legal
-        └── lib
-```
-
-To run the `jextract` tool, simply run the `jextract` command in the `bin` folder (again, the exact location of the binary might vary slightly depending on the platform, e.g. on macOs `build/jextract.app/Contents/MacOS/jextract` ):
+After building, there should be a new `jextract` folder under `build`.
+To run the `jextract` tool, simply run the `jextract` command in the `bin` folder:
 
 ```sh
 build/jextract/bin/jextract 
--- a/build.gradle
+++ b/build.gradle
@@ -4,10 +4,8 @@ import java.nio.file.Path;
 
 plugins {
     id "java"
-    // moditect for module-info.class injection and jlink
+    // moditect for module-info.class injection
     id "org.moditect.gradleplugin" version "1.0.0-rc3"
-    // jpackage plugin for packging jextract native app
-    id "org.panteleyev.jpackageplugin" version "1.3.1"
 }
 
 version = '1.0'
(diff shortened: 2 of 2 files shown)


mcimadamore (commented):
Looks good, the only nit is that I can see few JDK-specific launchers in the bin folder which can be confusing - e.g. `javac`, `java`, `serialver` etc. Would it be possible to drop them?

mcimadamore:
> Looks good, the only nit is that I can see few JDK-specific launchers in the bin folder which can be confusing - e.g. `javac`, `java`, `serialver` etc. Would it be possible to drop them?

Uhmmm - seems like jlink doesn't allow for that, and if we clean up all executables, then the jextract script fails because it can't find "java".

mcimadamore:
I believe README needs changes.
@overheadhunter, does this work with your plugin?

overheadhunter:
> @overheadhunter, does this work with your plugin?

So far one needs to manually specify the exact path of the jextract "binary" (launcher script should behave the same) anyway. I don't expect any problems. If there are, I'll adjust the plugin, so please go ahead! 🙂

mcimadamore:
> > @overheadhunter, does this work with your plugin?
> 
> So far one needs to manually specify the exact path of the jextract "binary" (launcher script should behave the same) anyway. I don't expect any problems. If there are, I'll adjust the plugin, so please go ahead! slightly_smiling_face

Ok, thanks. What this change does, is that it makes the build output more regular, so your plugin can just take a variable that points to "jextract home".

README.md
```diff
@@ -26,22 +26,8 @@ $ sh ./gradlew -Pjdk18_home=<jdk18_home_dir> -Pllvm_home=<libclang_dir> clean ve
...
-        ├── bin
-        ├── conf
-        ├── include
-        ├── legal
-        └── lib
-```
-
-To run the `jextract` tool, simply run the `jextract` command in the `bin` folder (again, the exact location of the binary might vary slightly depending on the platform, e.g. on macOs `build/jextract.app/Contents/MacOS/jextract` ):
+After building, there should be a new `jextract` folder under `build`.
+To run the `jextract` tool, simply run the `jextract` command in the `bin` folder:
```
mcimadamore, line 30: should say `jextract.bat` for windows (e.g. in parens) ?

sundararajana: Hmm... "jextract' command is what we say. On Windows jextract.exe, jextract.bat both will be called "jextract" command (because user doesn't run with the extension), right?

mcimadamore (approved):

README.md
sundararajana: /integrate

PR #14, opened 2022-04-06
7903146: TypeTranslator is no longer needed

mcimadamore:
Now that ValueLayout has carrier information attached, we no longer need to perform two separate visit on a function type, one to get a FunctionDescriptor and one to get a MethodType. We can just get a FunctionDescriptor and derive everything from there.

This patch applies several cleanups on OutputFactory, to flatten the code, and remove some spaghetti lambdas. It also drops the FunctionInfo/VarInfo abstraction, since now that we can derive most of the info from layouts, these do not seem to add much value.


src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -272,9 +277,12 @@ private String getConstantString(Class<?> type, Object value) {
                 buf.append("d");
             } else {
                 buf.append("Double.valueOf(\"");
-                buf.append(value.toString());
+                buf.append(value);
                 buf.append("\")");
             }
+        } else if (type == boolean.class) {
```
mcimadamore, line 283: This is a bug that was revealed by the cleanup: for boolean constants macros, we were using a "boolean" layout - but then we were dropping that information away, and just using byte as a carrier. That's because the Type -> Layout translation was subtly different from the Type -> Class one. Now that we only use a single translation, all these discrepancies are exposed.

sundararajana (approved):
LGTM

PR #17, opened 2022-04-07
Re-arrange test directory structure

mcimadamore:
Looking more at jextract tests, when we moved things in the standalone repo, few things went wrong. For instance, `TestSplit` cannot be executed right now, because it would need to be ran with some custom JDK properties. But since it's now part of a testng test structure, that is not possible.

We'd need to move TestSplit in a different folder, but if we do that we have a problem, in that TestSplit depends on the TestUtils and JextractToolRunner helpers.

This patch moves all the helper classes (JextractToolRunner, TestUtils, JextractAPITestBase) in the `testlib` package under the `test/lib` folder.

Then, this patch restructures the test folder, as follows:

* the `java` folder is renamed to `testng` - all testng test w/o jtreg header live here
* the `generator` folder is moved under a new folder called `jtreg`; under this new folder we find tests with regular jtreg header

With all this, we can now add a TEST.properties file under `test/testng` which contains a pointer to the lib folder that should be used by testng tests.

And, we can finally move TestSplit under `test/jtreg`, and add the jtreg test header we need to be able to run this correctly.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jextract pull/17/head:pull/17` \
`$ git checkout pull/17`

`$ git checkout pull/17` \
`$ git pull https://git.openjdk.java.net/jextract pull/17/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 17`

`$ git pr show -t 17`

</details>

<a href="https://git.openjdk.java.net/jextract/pull/17.diff">https://git.openjdk.java.net/jextract/pull/17.diff</a>

</details>


test/lib/testlib/JextractToolRunner.java
```diff
@@ -91,10 +87,9 @@ protected JextractToolRunner() {
 
     protected JextractToolRunner(Path input, Path output) {
          inputDir = (input != null) ? input :
-                Paths.get(System.getProperty("test.root", "."),
-                        "java", "org", "openjdk", "jextract", "test", "toolprovider");
+                Paths.get(System.getProperty("test.file")).getParent();
```
mcimadamore, line 90: This path is now independent from where tests are located

test/lib/testlib/JextractApiTestBase.java
```diff
@@ -36,14 +36,13 @@
...
+    protected static final boolean isMacOSX =
             System.getProperty("os.name", "unknown").contains("OS X");
-    static final boolean isWindows =
+    protected static final boolean isWindows =
             System.getProperty("os.name", "unknown").startsWith("Windows");
 
     public static  Declaration.Scoped parse(String headerFilename, String... parseOptions) {
-        Path header = Paths.get(System.getProperty("test.root", "."),
-                "java", "org", "openjdk", "jextract", "test", "api", headerFilename);
+        Path header = Paths.get(System.getProperty("test.file")).getParent().resolve(headerFilename);
```
mcimadamore, line 45: This path is now independent from where tests are located

sundararajana:
latest change looks good to me.

PR #18, opened 2022-04-11
7903154: jextract build fails on Windows

sundararajana:
Apart from quote issue for --add-options, @mcimadamore found that the verify target failed as well. "jextract.bat" has to be launched on Windows (instead of "jextract")


mcimadamore (approved):
clean/verify works well on Windows

PR #25, opened 2022-04-16
gradle buildscript improvements

danieljarabek:
Some improvements based on suggestions I made in an email [1]. Since then, the buildscript has a evolved quite a bit so some of my suggestions became obsolete or easier. This PR generally implements suggestions 2 and 3 from my email.

Use Gradle incremental builds and up to date checks:
- The `createJextractImage` and `createRuntimeImageForTest` tasks now use Gradle task input and ouputs to enable up to date checks.
- The cmake tasks could theoretically also have inputs and outputs added, however cmake has incremental builds already and these tasks are only run when running jtreg and take little time compared to the test suite.

I made a few minor style changes and fixed the comment on the jtreg task.

I would like feedback on one change, which is making the assemble task automatically run the jextractapp task to build a jextract binary. Do you think it is best if the assemble task automatically runs the verify task as well? Or should things just be left as they are and require running verify or jextractapp manually? For some brief insight into Gradle lifecycle tasks for those not familiar: The build task should build and test everything so it depends on assemble and check. The assemble task is responsible for building everything, it depends on jar in a default configuration. The check task is responsible for running tests. For more information see the Gradle docs [2].

[1] - https://mail.openjdk.java.net/pipermail/panama-dev/2022-March/016665.html
[2] - https://docs.gradle.org/current/userguide/base_plugin.html#sec:base_tasks

---------
### Progress
- [x] Change must be properly reviewed (no reviews required)

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.java.net/census#mcimadamore) (@mcimadamore - Committer) ⚠️ Review applies to [7d71f4a8](https://git.openjdk.java.net/jextract/pull/25/files/7d71f4a855b10597ca5ff704031aaf379df468b5)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jextract pull/25/head:pull/25` \
`$ git checkout pull/25`

`$ git checkout pull/25` \
`$ git pull https://git.openjdk.java.net/jextract pull/25/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 25`

`$ git pr show -t 25`

</details>

<a href="https://git.openjdk.java.net/jextract/pull/25.diff">https://git.openjdk.java.net/jextract/pull/25.diff</a>

</details>


build.gradle
```diff
@@ -27,9 +27,6 @@ def clang_version = clang_versions[0]
 def jextract_app_dir = "$buildDir/jextract"
 def clang_include_dir = "${llvm_home}/lib/clang/${clang_version}/include"
 checkPath(clang_include_dir)
-def quote_jlink_opts = Os.isFamily(Os.FAMILY_WINDOWS)?
-        "\\\"--enable-native-access=org.openjdk.jextract\\\"" :
-        '"--enable-native-access=org.openjdk.jextract"';
```
sundararajana, line 32: global moved to place of use. looks good.

build.gradle
```diff
@@ -40,7 +37,7 @@ repositories {
 }
 
 compileJava {
-    options.compilerArgs.addAll(['--release', '18'])
+    options.release = 18
```
sundararajana, line 40: using property style rather passing option as string. looks fine

build.gradle
```diff
@@ -55,7 +52,17 @@ jar {
...
+    inputs.file(jar.archiveFile.get())
+    outputs.dir(jextract_app_dir)
+
+    def quote_jlink_opts = Os.isFamily(Os.FAMILY_WINDOWS)?
+        '\\"--enable-native-access=org.openjdk.jextract\\"' :
+        '"--enable-native-access=org.openjdk.jextract"'
+
+    doFirst {
+        delete(jextract_app_dir)
+    }
```
sundararajana, line 65: I didn't know about inputs/outputs of Exec task. I just checked that running verify without clean in between does not rebuild (after your change). So looks fine.

build.gradle
```diff
@@ -150,7 +169,7 @@ task cmakeBuild(type: Exec) {
     ]
 }
 
-// run jtreg tests. Note: needs jtreg_cmd variable set to point to the jtreg
+// run jtreg tests. Note: needs jtreg_home variable set to point to the jtreg
```
sundararajana, line 172: good catch :) That comment is relic of the past.

build.gradle
```diff
@@ -102,12 +112,21 @@ task verify(type: Exec) {
...
-    onlyIf { !new File("$buildDir/jextract-jdk-test-image").exists() }
+    def out_dir = "$buildDir/jextract-jdk-test-image"
+
+    // if these inputs or outputs change, gradle will rerun the task
+    inputs.file(jar.archiveFile.get())
+    outputs.dir(out_dir)
+
+    doFirst {
+        delete(out_dir)
+    }
```
sundararajana, line 123: similar inputs/outputs change. looks good.

build.gradle
```diff
@@ -90,6 +97,9 @@ task jextractapp() {
     dependsOn copyLibClang
 }
 
+// build the jextract image when the build or assemble task is run
+assemble.dependsOn(jextractapp)
+
```
sundararajana, line 102: Not sure of this change. Basically what we want is only three targets to use "clean", "verify" and "jtreg"

"verify" is for just create minimal jextract JDK image and run simple 'integration test' on it.
'jtreg' is for really running all tests

Note that we cannot use usual lifecycle of Gradle projects. Out tests are run a separate JDK image containing jextract tool in it - because this is how we expect users to deploy/use.

"verify" creates a minimal jextract image (which won't have compiler and other things needed for jtreg test run and so we create two JRE images) and runs a simple jextract on a simple test.h.

build.gradle
sundararajana: I am not a committer in this project. So, we've to wait for @mcimadamore for sponsoring the change anyway. I'd like him to review and sponsor. Thanks for your contribution @danieljarabek !

build.gradle
sundararajana: "jtreg" target requires additional toolset (cmake, C compiler, jtreg) at the user end. This is not the usual pure Java project "tests" which don't need any of these. Already "verify" doesn't depend on "jtreg" (verify is just build of jextract image + a simple integration test to extract a test.h). "jtreg" perhaps can skip the "project jextract image" + "integration test" step. i.e., after jar straight build test JDK image with jextract and run all jextract tests. But, not sure how much is saved by that.

danieljarabek: To me at least, it would make sense for the "build" task to create the final artifact designed for users. This help would avoid confusion from people who didn't fully read the README, and try to simply run the "build" task (then try to run the jar in build/libs or otherwise do something incorrect/unsupported). Perhaps it may even be best to print a message announcing the correct location of the final artifact since developers familiar with gradle may assume that it just builds a executable jar like most projects. However I also understand how this project doesn't fully fit into the gradle build lifecycle. I'll leave it up to you two to decide what we want the final build process to look like. I'm happy to implement whatever approach you decide on in this PR.

mcimadamore: At the moment it is:

build -> build the jextract classes
verify -> build jextract image and run small integration test - depends on build
jtreg -> run all jtreg tests - depends on verify

Now, we don't want "jtreg" be part of the build pipeline, because running tests require setting up compilation of native libraries, which might not be possible for users. 

That said, on "verify" we're more flexible. I think another good stacking could be:

build -> build the jextract classes and image
verify -> run small integration test - depends on build
jtreg -> run all jtreg tests - depends on verify

This means that doing `gradle build` (which is probably what majority of users will reach out for first) will produce something useful.

danieljarabek: This change currently implements the second stacking you propose (with the minor detail of assemble depending on jextractapp since build depends on assemble by default). Is this acceptable or should I revert this change?

mcimadamore: I've tried out this patch, and I confirm that it behaves as described (e.g. second stacking). I had an offline discussion with @sundararajana who is also ok with this change. This can go ahead IMHO. I will approve.

danieljarabek:
/integrate

Thank You! I hope to make more contributions in the future to jextract and other OpenJDK projects.

PR #39, opened 2022-05-25
7903200: sync makefiles to build jmod for jextract

sundararajana:
sync makefiles for jmod creation (and update for jdk19). renamed jmod file name


mcimadamore (approved):
Looks good

PR #45, opened 2022-05-30
7903202: Enable macOS in Github Action CI config

denismakogon:
This adjustment modifies the build config by introducing
strategy matrix in order to enable builds from both Ubuntu and macOS.

---------
### Progress

### Issue
 * [CODETOOLS-7903202](https://bugs.openjdk.org/browse/CODETOOLS-7903202): Enable macOS in Github Action CI config

### Reviewers
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - no project role)
 * [Athijegannathan Sundararajan](https://openjdk.java.net/census#sundar) (@sundararajana - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jextract pull/45/head:pull/45` \
`$ git checkout pull/45`

`$ git checkout pull/45` \
`$ git pull https://git.openjdk.java.net/jextract pull/45/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 45`

`$ git pr show -t 45`

</details>

<a href="https://git.openjdk.java.net/jextract/pull/45.diff">https://git.openjdk.java.net/jextract/pull/45.diff</a>

</details>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ env:
 
 jobs:
 
-  linux-x64:
+  build-and-test:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,10 +7,24 @@ on:
       - pr/*
   workflow_dispatch:
 
+env:
+  CLANG_LLVM_BASE_URL: "https://github.com/llvm/llvm-project/releases/download/llvmorg-13.0.0/clang+llvm-13.0.0-x86_64"
+  ARCHIVE_EXT: "tar.xz"
+
 jobs:
 
   linux-x64:
-    runs-on: "ubuntu-20.04"
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-latest, ubuntu-20.04]
+        include:
+          - os: ubuntu-20.04
+            TARGET: linux-gnu-ubuntu-20.04
+            JAVA19_HOME: /tmp/deps/jdk-19
+          - os: macos-latest
+            TARGET: apple-darwin
+            JAVA19_HOME: /tmp/deps/jdk-19/jdk-19.jdk/Contents/Home
 
     steps:
     - name: 'Check out repository'
@@ -21,7 +35,8 @@ jobs:
     - name: 'Prepare'
       shell: sh
       run: |
-        mkdir deps
+        mkdir -p deps/jtreg
+        mkdir -p /tmp/deps
 
     - name: 'Download JDK 19'
       id: download_jdk_19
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -21,7 +35,8 @@ jobs:
     - name: 'Prepare'
       shell: sh
       run: |
-        mkdir deps
+        mkdir -p deps/jtreg
+        mkdir -p /tmp/deps
```
JornVernee, line 39: Looks like you moved `deps` to `/tmp`. Why is that?

Also, should the `deps/jtreg` also be moved to `tmp` in that case?

.github/workflows/test.yml
```diff
@@ -7,10 +7,24 @@ on:
...
+    strategy:
+      matrix:
+        os: [macos-latest, ubuntu-20.04]
+        include:
+          - os: ubuntu-20.04
+            TARGET: linux-gnu-ubuntu-20.04
+            JAVA19_HOME: /tmp/deps/jdk-19
+          - os: macos-latest
+            TARGET: apple-darwin
+            JAVA19_HOME: /tmp/deps/jdk-19/jdk-19.jdk/Contents/Home
```
JornVernee, line 27: Both of these test runs now show up as part of the same job called `linux-x64`. I suggest splitting this into 3 jobs instead:
1. a job that does all the setup.
2. a job for linux testing
3. a job for mac testing

Where both 2 and 3 depend on 1.  (see for instance how the JDK workflow achieves this using job outputs: https://github.com/openjdk/jdk/blob/master/.github/workflows/submit.yml#L149)

'Testing' would include the 'Build Jextract' and 'Run Tests' steps.

JornVernee: Actually, it looks like that might not be that easy, since the prerequisites also have to pick a platform to run on.

In that case, I suggest just changing the name of the job to be something like `Build & Test`.

.github/workflows/test.yml
denismakogon: there was a problem with relative paths on different platforms, so, replacing relative `.deps` to `/tmp/deps` works and feels better, the only leftover is `jtreg` because `action/checkout` can't checkout a project into a temp folder.

.github/workflows/test.yml
denismakogon: At first, i thought about replacing one config with per-platform. The only foreseeable scenario with a per-platform config - is when we'd need to add Windows. Because eventually, we'd have to release a binary for Win as well. But as of now, one config for both Debian and macOS is a good-to-go solution.

denismakogon: okay, I renamed a job to `build-and-test`

JornVernee (approved):
This looks good to me. Just have to wait for the OCA check, you might want to ping Dalibor for that.

Also, I think you actually want `/covered` instead of `/signed`.

mcimadamore (approved):
Looks good

sundararajana (approved):
LGTM

PR #58, opened 2022-07-29
7903239: ofAddress factory of function pointer type is wrong for struct returns

lost-illusi0n:
When generating the lambda inside the `ofAddress` factory, it is missing the `SegmentAllocator` needed when the function returns a struct. This is [implemented for regular functions](https://github.com/openjdk/jextract/blob/0582eaf1b4cdba95f0ee8c2480767433bb647d0d/src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java#L122-L127), but not for function pointers.

---------
### Progress

### Issue
 * [CODETOOLS-7903239](https://bugs.openjdk.org/browse/CODETOOLS-7903239): ofAddress factory of function pointer type is wrong for struct returns (**Bug** - P3) ⚠️ Issue is already resolved. Consider making this a "backport pull request" by setting the PR title to `Backport <hash>` with the hash of the original commit. See [Backports](https://wiki.openjdk.org/display/SKARA/Backports).

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - Committer) 🔄 Re-review required (review was made when pull request targeted the [master](https://git.openjdk.org/jextract/tree/master) branch)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/58/head:pull/58` \
`$ git checkout pull/58`

`$ git checkout pull/58` \
`$ git pull https://git.openjdk.org/jextract.git pull/58/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 58`

`$ git pr show -t 58`

</details>

<a href="https://git.openjdk.org/jextract/pull/58.diff">https://git.openjdk.org/jextract/pull/58.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/58#issuecomment-1200137588)


test/testng/org/openjdk/jextract/test/toolprovider/Test7903239.java
```diff
@@ -0,0 +1,41 @@
...
+
+import testlib.TestUtils;
+import org.testng.annotations.Test;
+import testlib.JextractToolRunner;
+
+import java.nio.file.Path;
+
+public class Test7903239 extends JextractToolRunner {
+    @Test
+    public void testFunctionPointerReturningStruct() {
```
mcimadamore, line 34: While this test in itself is good, I feel it's perhaps the wrong kind of test, in the sense that it only checks that jextract completes normally. Maybe a test that uses the jextract jtreg extension would be better? For inspiration look here:

https://github.com/openjdk/jextract/blob/master/test/jtreg/generator/funcPointerInvokers/TestFuncPointerInvokers.java

You can basically run jextract as part of a jtreg test action, as here:

https://github.com/openjdk/jextract/blob/master/test/jtreg/generator/funcPointerInvokers/TestFuncPointerInvokers.java#L39

Then you can actually compile (against the jextracted bits) and even run:

https://github.com/openjdk/jextract/blob/master/test/jtreg/generator/funcPointerInvokers/TestFuncPointerInvokers.java#L40

So you can check that the generated code actually behaves as expected at runtime.

test/jtreg/generator/test7903239/libTest7903239.c
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
```
mcimadamore, line 2: If the test is new, should it have 2022 in the copyright year? (also in c file)

test/jtreg/generator/test7903239/Test7903239.java
```diff
@@ -0,0 +1,58 @@
...
+ * @run main/othervm JtregJextractSources -l Test7903239 -t test.jextract.test7903239 test7903239.h
+ * @run testng/othervm --enable-native-access=ALL-UNNAMED Test7903239
+ */
+public class Test7903239 {
+    @Test
+    public void testOfAddress() {
+        try (MemorySession session = MemorySession.openConfined()) {
+            var foo = test7903239_h.foo$SEGMENT();
+
+            var barA = Foo.a(foo, session).apply();
```
mcimadamore, line 51: I think the limitation of the strategy are clearer in this example :-)
The allocator is decided once and for all when the functional interface instance is obtained. Any other call to `apply` will use the same segment allocator specified on creation. To solve this, I believe the only way is to use the default method setup I described earlier. But the patch is good enough as is for now.

lost-illusi0n: Yes, the tradeoff is quite apparent, but I suppose reasonable to an extent. Your solution is probably the best road to go down when playing around with it though :+1:.

test/jtreg/generator/test7903239/libTest7903239.c
```diff
@@ -21,21 +21,14 @@
...
-    @Test
-    public void testFunctionPointerReturningStruct() {
-        Path outputPath = getOutputFilePath("7903239_gen");
-        Path headerFile = getInputFilePath("test7903239.h");
-        run("--output", outputPath.toString(), headerFile.toString()).checkSuccess();
-        // nothing is generated that we can check, so we just check that jextract ran successfully
-        TestUtils.deleteDir(outputPath);
-    }
-}
+struct Foo foo = { a, b };
```
mcimadamore, line 34: Watch out: newline needed at the end

mcimadamore (approved):

mcimadamore:
I have not sponsored this yet as the target branch of this PR seems wrong (e.g. `master` instead of `panama` or `jdk19`).

lost-illusi0n:
Woops, should be good now.

PR #66, opened 2022-08-30
7903271: Add sample of libzstd

minborg:
This PR adds a sample using `libzstd`, a high-performance compression library.

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Change must be properly reviewed (no review required)

### Issue
 * [CODETOOLS-7903271](https://bugs.openjdk.org/browse/CODETOOLS-7903271): Add sample of libzstd

### Reviewers
 * [Athijegannathan Sundararajan](https://openjdk.org/census#sundar) (@sundararajana - Committer)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract pull/66/head:pull/66` \
`$ git checkout pull/66`

Update a local copy of the PR: \
`$ git checkout pull/66` \
`$ git pull https://git.openjdk.org/jextract pull/66/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 66`

View PR using the GUI difftool: \
`$ git pr show -t 66`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jextract/pull/66.diff">https://git.openjdk.org/jextract/pull/66.diff</a>

</details>

PR #68, opened 2022-08-31
Add path independence

minborg:
This small PR incorporated feedback from the original PR: https://github.com/openjdk/jextract/pull/66 made after it was integrated.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/68/head:pull/68` \
`$ git checkout pull/68`

`$ git checkout pull/68` \
`$ git pull https://git.openjdk.org/jextract.git pull/68/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 68`

`$ git pr show -t 68`

</details>

<a href="https://git.openjdk.org/jextract/pull/68.diff">https://git.openjdk.org/jextract/pull/68.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/68#issuecomment-1232560695)


sundararajana (commented):
I guess we need similar change for many other samples. One flip side of the approach in this PR that we lose the specific version to use. C APIs need not be backward compatible (many are not!). Sample may work on a specific version of C library and not on the latest (or older) brew installed version.

minborg:
Perhaps we could mention a specific version in the README file (e.g. the example here is verified for foo version x.y.z) and that appropriate changes might have to be made if another version is used.

bondolo:
> I guess we need similar change for many other samples. One flip side of the approach in this PR that we lose the specific version to use. C APIs need not be backward compatible (many are not!). Sample may work on a specific version of C library and not on the latest (or older) brew installed version.

As long as the library remains compatible the sample should continue to work. It seemed wrong to install a non-specific version using homebrew (or apt or yum) and then expect that specific version had been installed.

I agree with the suggestion that the readme should document which version of the library was used.

bondolo (commented):
LGTM!

It is tempting to get fancy and implement everything needed to simultaneously support Linux and MacOS but the result would obscure the clarity of the sample so I agree to keep it simple.

(I am an [OpenJDK Panama](https://openjdk.org/census#mduigou) reviewer but have been inactive for 8+ years on the project so defer to more active contributors for approval).

sundararajana:
Note: "As long as the library remains compatible.." That is not true for all samples. For eg. libffmpeg has broken compatibility. Only specific older version works.  May be the PR title should "Add path independence for libzstd sample".

PR #74, opened 2022-09-21
Add Dockerfile, zero-dep reproducible builds

GavinRay97:
As title says, adds a Dockerfile that will fetch all necessary deps and build `jextract` for you on Ubuntu 22.10

It would be nice if some official JDK organization could put this on their Dockerhub, so that users could just do:

```sh
$ docker run --rm -it openjdk/jextract:latest
```

I've pushed it to my personal Dockerhub, so if anyone would like to test it out, you can use the README directions but prepend `gavinray/` before the image name:

```sh
$ echo "struct Foo { int a; long b; }; int use_foo(struct Foo foo);" >> sample.h

$ docker run --rm -it -v ${PWD}:/tmp/jextract \
    gavinray/jextract:latest \
    -I /tmp/jextract \
    --source \
    -l libfoo \
    --output /tmp/jextract/generated \
    --target-package com.example \
    /tmp/jextract/sample.h

$ tree ./generated
generated/
└── com
    └── example
        ├── Constants$root.java
        ├── Foo.java
        ├── RuntimeHelper.java
        ├── constants$0.java
        └── sample_h.java
```

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/74/head:pull/74` \
`$ git checkout pull/74`

`$ git checkout pull/74` \
`$ git pull https://git.openjdk.org/jextract.git pull/74/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 74`

`$ git pr show -t 74`

</details>

<a href="https://git.openjdk.org/jextract/pull/74.diff">https://git.openjdk.org/jextract/pull/74.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/74#issuecomment-1264033814)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,57 @@
+FROM ubuntu:22.10 AS builder
+
+# Allow configuring JDK version
+# COMPILE_VERSION will be used with the gradle build task
+# RUNTIME_VERSION will be passed to the "-Pjdk19_home" gradle property
+ARG SDKMAN_JAVA_COMPILE_VERSION=17-open
+ARG SDKMAN_JAVA_RUNTIME_VERSION=19-open
+
+# Allow configuring LLVM download URL
+ARG LLVM_DOWNLOAD_URL=https://github.com/llvm/llvm-project/releases/download/llvmorg-14.0.0/clang+llvm-14.0.0-x86_64-linux-gnu-ubuntu-18.04.tar.xz
+
+SHELL ["/bin/bash", "-c"]
+
+# Install dependencies
+RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get install -y \
+    git \
+    curl \
+    wget \
+    bash \
+    unzip \
+    zip \
+    xz-utils
+
+# Install sdkman
+RUN curl -s "https://get.sdkman.io" | bash
+
+# Use sdkman to install JDK 19
+RUN source "$HOME/.sdkman/bin/sdkman-init.sh" \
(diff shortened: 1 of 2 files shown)


GavinRay97:
I signed the Oracle OCA for contributing to GraalVM already =/

https://github.com/oracle/graal/pull/3603#issuecomment-889122954

image

JornVernee:
@GavinRay97 I see you in the OCA signatories list with the graal project listed. I'm not sure why the bot isn't picking this up. I'll ask around.

JornVernee:
Hi @GavinRay97 sorry for the late reply.

After some internal discussions, we have decided not to take on this patch, or any other that adds a docker file.

A docker file pulls in several dependencies from the internet, which have to be individually vetted and approved on a per-version basis.

As we already do this for the gradle build, we are wary of adding more process overhead by adding a docker file (or other things that pull in more dependencies) to the repo as well. Besides that, it's another build configuration to maintain, and we already have 2 (gradle and make).

GavinRay97:
That's fair enough, maybe I could redo the PR where instead of putting the
Dockerfile in the repo,
an example definition that users can use for their own Dockerfiles if they
wish to build jextract
is provided in the documentation?

The problem I was hoping to solve is preventing users from needing to
manually install and
configure things like the right version of the LLVM toolchain for their OS,
etc

Would this be an acceptable compromise? Or is something like this still
better off in third-party land?

JornVernee:
I think that would suffer from the same issues I'm afraid.

PR #85, opened 2022-10-12
7903257: jextract should generate javadoc comments

sundararajana:
Generating C signature as doc comments.

---------
### Progress

### Issue
 * [CODETOOLS-7903257](https://bugs.openjdk.org/browse/CODETOOLS-7903257): jextract should generate javadoc comments

### Reviewers
 * @minborg (no known github.com user name / role) ⚠️ Review applies to [deb57bf1](https://git.openjdk.org/jextract/pull/85/files/deb57bf1bc301ea4286b1ac1f9a17269be00feb3)
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - no project role) ⚠️ Review applies to [3af5a8cc](https://git.openjdk.org/jextract/pull/85/files/3af5a8cc51013881f89623f8d94d6ca9ac8ffeac)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract pull/85/head:pull/85` \
`$ git checkout pull/85`

`$ git checkout pull/85` \
`$ git pull https://git.openjdk.org/jextract pull/85/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 85`

`$ git pr show -t 85`

</details>

<a href="https://git.openjdk.org/jextract/pull/85.diff">https://git.openjdk.org/jextract/pull/85.diff</a>

</details>

### DIFF
--- a/src/main/java/org/openjdk/jextract/impl/CDeclarationPrinter.java
+++ b/src/main/java/org/openjdk/jextract/impl/CDeclarationPrinter.java
@@ -197,6 +197,8 @@ final class CDeclarationPrinter implements Declaration.Visitor<Void, Void> {
                     return prefixedType("signed", t);
                 case VOLATILE:
                     return prefixedType("volatile", t);
+                case COMPLEX:
+                    return prefixedType("complex", t);
                 case default:
                     return new TypeVisitorResult(false, t.name().get());
             }
--- a/src/main/java/org/openjdk/jextract/impl/CDeclarationPrinter.java
+++ b/src/main/java/org/openjdk/jextract/impl/CDeclarationPrinter.java
@@ -29,6 +29,7 @@ package org.openjdk.jextract.impl;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.stream.Collectors;
 import org.openjdk.jextract.Declaration;
 import org.openjdk.jextract.Type;
@@ -102,7 +103,7 @@ final class CDeclarationPrinter implements Declaration.Visitor<Void, Void> {
         indent();
         var ftype = d.type();
         var rtype = ftype.returnType();
-        builder.append(nameAndType(rtype, ""));
+        builder.append(typeStr(rtype));
         builder.append(" ");
         builder.append(d.name());
         builder.append("(");
@@ -130,15 +131,24 @@ final class CDeclarationPrinter implements Declaration.Visitor<Void, Void> {
     @Override
     public Void visitConstant(Declaration.Constant d, Void ignored) {
         indent();
-        builder.append(nameAndType(d.type(), d.name()));
-        builder.append(" = ");
-        Object value = d.value();
-        if (value instanceof String str) {
-            builder.append("\"" + Utils.quote(str) + "\"");
+        Optional<String> enumName = EnumConstantLifter.enumName(d);
+        if (enumName.isPresent()) {
+            builder.append("enum " + enumName.get() + "." + d.name());
+            builder.append(" = ");
+            builder.append(d.value());
+            builder.append(";\n");
         } else {
-            builder.append(value);
+            builder.append("#define ");
+            builder.append(d.name());
+            Object value = d.value();
+            builder.append(" ");
+            if (value instanceof String str) {
+                builder.append("\"" + Utils.quote(str) + "\"");
+            } else {
+                builder.append(value);
+            }
+            builder.append("\n");
         }
-        builder.append(";\n");
         return null;
     }
 
@@ -166,10 +176,19 @@ final class CDeclarationPrinter implements Declaration.Visitor<Void, Void> {
             typeStr : (typeStr + " " + name);
     }
 
+    private static String typeStr(Type type) {
+        var result = type.accept(typeVisitor, "");
+        assert !result.nameIncluded();
+        return result.typeStr();
+    }
+
     // result type for Type.Visitor
     private record TypeVisitorResult(boolean nameIncluded, String typeStr) {}
 
     private static Type.Visitor<TypeVisitorResult, String> typeVisitor = new Type.Visitor<>() {
+        // context argument in this visitor starts with a name. But it may pick up
+        // "*" prefixes for pointer type.
+
         @Override
         public TypeVisitorResult visitPrimitive(Type.Primitive t, String context) {
             return new TypeVisitorResult(false, t.kind().typeName());
--- a/src/main/java/org/openjdk/jextract/impl/CDeclarationPrinter.java
+++ b/src/main/java/org/openjdk/jextract/impl/CDeclarationPrinter.java
@@ -1,0 +1,257 @@
+/*
+ *  Copyright (c) 2022, Oracle and/or its affiliates. All rights reserved.
+ *  DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ *  This code is free software; you can redistribute it and/or modify it
+ *  under the terms of the GNU General Public License version 2 only, as
+ *  published by the Free Software Foundation.  Oracle designates this
+ *  particular file as subject to the "Classpath" exception as provided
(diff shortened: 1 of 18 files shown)


minborg (approved):

src/main/java/org/openjdk/jextract/impl/CDeclarationPrinter.java
```diff
@@ -0,0 +1,261 @@
...
+
+    // In few cases, C type signature 'embeds' name.
+    // Examples:
+    //     int a[3]; // 'a' in between int and []
+    //     int (*func)(int); // 'func' is inside paren after '*'
+    // TypeVisitor accepts name and includes it in the appropriate
+    // place as needed. If not included, boolean flag is set to false
+    // in the result.
+
+    private static String nameAndType(Type type, String name) {
```
mcimadamore, line 162: Maybe we should call this "varDecl" since this is morally to be used for variable declarations.

src/main/java/org/openjdk/jextract/impl/CDeclarationPrinter.java
```diff
@@ -0,0 +1,261 @@
...
+        }
+        return null;
+    }
+
+    @Override
+    public Void visitFunction(Declaration.Function d, Void ignored) {
+        indent();
+        var ftype = d.type();
+        var rtype = ftype.returnType();
+        builder.append(nameAndType(rtype, ""));
```
mcimadamore, line 105: I would rather add another helper function to just print a type, rather than reusing `nameAndType` here. Also, we should check that the type doesn't have a name attached.

src/main/java/org/openjdk/jextract/impl/CDeclarationPrinter.java
mcimadamore: It's also not super clear to me what happens when the type has a name, and the user also specified a name. Can that ever happen? Or, if this happens, both names should be the same (in which case we should assert that).

src/main/java/org/openjdk/jextract/impl/CDeclarationPrinter.java
```diff
@@ -0,0 +1,261 @@
...
+        }
+        builder.append(";\n");
+        return null;
+    }
+
+    @Override
+    public Void visitTypedef(Declaration.Typedef d, Void ignored) {
+        indent();
+        builder.append("typedef ");
+        builder.append(nameAndType(d.type(), d.name()));
```
mcimadamore, line 149: This use of `nameAndType` also leaves me a bit doubtful. There's var declarations, and there's just types + name. What's acceptable in both cases is different - e.g. `int a[3]` is not a valid type for a typedef. So, in cases where we just want a type, and the type should not have a name, we should have a way to say so in the code.

src/main/java/org/openjdk/jextract/impl/CDeclarationPrinter.java
```diff
@@ -0,0 +1,261 @@
...
+        public TypeVisitorResult visitDeclared(Type.Declared t, String context) {
+            Declaration.Scoped scoped = t.tree();
+            return new TypeVisitorResult(false, typeTag(scoped) + " " + scoped.name());
+        }
+
+        @Override
+        public TypeVisitorResult visitArray(Type.Array t, String context) {
+            String brackets = String.format(" %s[%s]", context,
+                t.elementCount().isPresent() ? t.elementCount().getAsLong() : "");
+            return new TypeVisitorResult(true, t.elementType().accept(this, "").typeStr() + brackets);
```
mcimadamore, line 244: Is there always a name attached to an array type? I would say no? (e.g. in a typedef)

src/main/java/org/openjdk/jextract/impl/CDeclarationPrinter.java
```diff
@@ -0,0 +1,261 @@
...
+                }
+                argsStr = args.stream()
+                    .collect(Collectors.joining(",", "(", ")"));
+            } else {
+                argsStr = t.argumentTypes().stream()
+                    .map(a -> a.accept(this, "").typeStr())
+                    .collect(Collectors.joining(",", "(", ")"));
+            }
+            String res = t.returnType().accept(this, "").typeStr();
+            return new TypeVisitorResult(true, res + " (" + context + ")" + argsStr);
```
mcimadamore, line 231: Is there always a name attached to a function pointer? I'd say no (e.g. typedef).

src/main/java/org/openjdk/jextract/impl/CDeclarationPrinter.java
sundararajana: Example:

typedef int (*handler)(int x);

The type signature has buried name here ("handler")

src/main/java/org/openjdk/jextract/impl/CDeclarationPrinter.java
sundararajana: You're asking about parameter names? That's not present always. Hence the conditional code above.

src/main/java/org/openjdk/jextract/impl/CDeclarationPrinter.java
sundararajana: context is either empty string or has a name.

src/main/java/org/openjdk/jextract/impl/StructBuilder.java
```diff
@@ -48,15 +48,17 @@ class StructBuilder extends ConstantBuilder {
 
     private static final String MEMBER_MODS = "public static";
 
+    private final Declaration.Scoped structTree;
     private final GroupLayout structLayout;
     private final Type structType;
     private final Deque<String> prefixElementNames;
 
-    StructBuilder(JavaSourceBuilder enclosing, Declaration.Scoped tree,
+    StructBuilder(JavaSourceBuilder enclosing, Declaration.Scoped structTree,
```
mcimadamore, line 56: Should this class override `addVar` so that doc comments for getters and setters are also generated (like in the case of global vars) ?

mcimadamore (approved):
Looks good

src/main/java/org/openjdk/jextract/impl/StructBuilder.java
sundararajana: will check out.

src/main/java/org/openjdk/jextract/impl/CDeclarationPrinter.java
```diff
@@ -0,0 +1,280 @@
...
+        @Override
+        public TypeVisitorResult visitArray(Type.Array t, String context) {
+            String brackets = String.format(" %s[%s]", context,
+                t.elementCount().isPresent() ? t.elementCount().getAsLong() : "");
+            return new TypeVisitorResult(true, t.elementType().accept(this, "").typeStr() + brackets);
+        }
+
+        @Override
+        public TypeVisitorResult visitType(Type t, String context) {
+            return new TypeVisitorResult(false, t.getClass().getName());
```
JornVernee, line 268: Above in visitDelegated the default name seems to be `t.toString`. Maybe these could be factored out in some `defaultNameFor(Type type)` helper method for clarity.

mcimadamore (approved):

JornVernee (approved):

src/main/java/org/openjdk/jextract/impl/CDeclarationPrinter.java
```diff
@@ -0,0 +1,280 @@
...
+                case SIGNED:
+                    return prefixedType("signed", t);
+                case VOLATILE:
+                    return prefixedType("volatile", t);
+                case COMPLEX:
+                    return prefixedType("complex", t);
+                case default:
+                    // defensive. If no name is present, we don't want to crash
+                    return new TypeVisitorResult(false,
+                        t.name().isPresent()? t.name().get() : t.toString());
```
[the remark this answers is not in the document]
sundararajana: oops. missed your comments. Will piggyback in a later change.

PR #90, opened 2022-10-18
7903360: jextract does not handle function pointer return type properly

sundararajana:
* generating functional interface for pointer to function return types
* taking care of javadoc comment by using nameAndType function uniformly

---------
### Progress

### Issue
 * [CODETOOLS-7903360](https://bugs.openjdk.org/browse/CODETOOLS-7903360): jextract does not handle function pointer return type properly

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract pull/90/head:pull/90` \
`$ git checkout pull/90`

`$ git checkout pull/90` \
`$ git pull https://git.openjdk.org/jextract pull/90/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 90`

`$ git pr show -t 90`

</details>

<a href="https://git.openjdk.org/jextract/pull/90.diff">https://git.openjdk.org/jextract/pull/90.diff</a>

</details>

### DIFF
--- a/src/main/java/org/openjdk/jextract/impl/NameMangler.java
+++ b/src/main/java/org/openjdk/jextract/impl/NameMangler.java
@@ -130,6 +130,11 @@ final class NameMangler implements Declaration.Visitor<Void, Declaration> {
         return Objects.requireNonNull(declFiNames.get(nameAndDecl));
     }
 
+    String getReturnFiName(Declaration.Function func) {
+        Objects.requireNonNull(func);
+        return funcReturnID(func);
+    }
+
     String getFiName(Declaration parent, Declaration decl) {
         Objects.requireNonNull(decl);
         if (declFiNames.containsKey(decl)) {
@@ -142,6 +147,10 @@ final class NameMangler implements Declaration.Visitor<Void, Declaration> {
     }
 
     // Internals below this point
+
+    private static String funcReturnID(Declaration.Function func) {
+        return func.name() + "$return";
+    }
 
     private static String funcParamID(Declaration.Function func, int paramNum) {
         return func.name() + "$" + paramNum;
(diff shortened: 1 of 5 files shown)


src/main/java/org/openjdk/jextract/impl/NameMangler.java
```diff
@@ -130,6 +130,11 @@ String getFiName(Declaration.Function func, int paramNum, Declaration.Variable p
         return Objects.requireNonNull(declFiNames.get(nameAndDecl));
     }
 
+    String getReturnFiName(Declaration.Function func) {
```
mcimadamore, line 133: Why can't `getFiName` be reused here?

src/main/java/org/openjdk/jextract/impl/OutputFactory.java
```diff
@@ -258,6 +259,14 @@ public Void visitFunction(Declaration.Function funcTree, Declaration parent) {
             }
         }
 
+        // return type could be a function pointer type
+        Type.Function returnFunc = Utils.getAsFunctionPointer(funcTree.type().returnType());
```
mcimadamore, line 263: Did you check the corner case of a function pointer having a function pointer as its return type?

sundararajana: While we recurse for pointer to function parameter types, we did recurse for return type. That's being fixed here.

For pointer to functions, we didn't recurse even for arguments (and also for return types).

For example, for

void (*func)(void (*f)(int));


we generate FI for "func" but not for "f".

The current fix case is seen in standard library (signal in signal.h). We can deal with recursing args and return types for function pointer types in a separate fix.

src/main/java/org/openjdk/jextract/impl/NameMangler.java
mcimadamore: I missed that there were (already) two variants of getFIName. Ugh.

mcimadamore (approved):

PR #101, opened 2023-01-30
Fix jextract after removal of support for sub-byte layouts

mcimadamore:
This patch fixes some test failures which are caused by https://git.openjdk.org/panama-foreign/pull/766.

As that patch removes support for sub-byte layouts, we can no longer represent bitfields using group layouts (because we can no longer model group elements whose size is not a multiple off 8).

In reality, jextract isn't doing much with bitfields: it just generates a bunch of padding layouts of the right size and storing them into a group layout, but this layout, of course, cannot be used for access (because sub-byte layout doesn't work with VarHandles).

To fix this, I have added a new subtype of `Declaration.Variable` - namely, `Declaration.Bitfield` - this is like a variable - it has a name (but no layout), and has a type. On top of this, a `Bitfield` also stores an `offset` (relative to the contained) and a `width` (the size in bits of the bitfield).

Most of the changes in this patch are caused by the fact that the underlying IR for bitfields has changed (as described above). Since bitfields are still variables (whose kind is `BITFIELD`) and since variable of that kind are ignored by the jextract backend, nothing is going to change, really. That said, should we want to add bitfield accessors one day, all the info to generate the accessor should be available in the jextract IR.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract pull/101/head:pull/101` \
`$ git checkout pull/101`

`$ git checkout pull/101` \
`$ git pull https://git.openjdk.org/jextract pull/101/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 101`

`$ git pr show -t 101`

</details>

<a href="https://git.openjdk.org/jextract/pull/101.diff">https://git.openjdk.org/jextract/pull/101.diff</a>

</details>

### DIFF
--- a/src/main/java/org/openjdk/jextract/Declaration.java
+++ b/src/main/java/org/openjdk/jextract/Declaration.java
@@ -248,6 +248,22 @@ public interface Declaration {
     }
 
     /**
+     * A bitfield declaration. Same as a variable declaration, but doesn't have a layout. Instead, it has
+     * an offset (relative to the enclosing container) and a width.
+     */
+    interface Bitfield extends Variable {
+        /**
+         * {@return The bitfield offset (relative to the enclosing container)}
+         */
+        long offset();
+
+        /**
+         * {@return The bitfield width (in bits)}
+         */
+        long width();
+    }
+
+    /**
      * A constant value declaration.
      */
     interface Constant extends Declaration {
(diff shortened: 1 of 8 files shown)


sundararajana (approved):
LGTM

rmaucher:
Padding is there for the bitfields now (and most importantly without any error). Thanks !

PR #114, opened 2023-03-22
Bring jdk20 branch in sync with panama branch

mcimadamore:
This PR brings fixes from `panama` branch back to the `jdk20` branch.
We are doing this as part of getting ready to start shipping binary snapshot of the jdk20 version of jextract.

### DIFF
--- a/src/main/java/org/openjdk/jextract/impl/StructLayoutComputer.java
+++ b/src/main/java/org/openjdk/jextract/impl/StructLayoutComputer.java
@@ -154,6 +154,10 @@ final class StructLayoutComputer extends RecordLayoutComputer {
             if (!prevBitfieldDecls.isEmpty()) {
                 addField(bitfield(prevBitfieldDecls.toArray(new Declaration.Variable[0])));
             }
+            if (prevBitfieldSize == 0) {
+                // nothing to do
+                return;
+            }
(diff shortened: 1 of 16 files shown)


src/main/java/org/openjdk/jextract/impl/StructLayoutComputer.java
```diff
@@ -148,19 +140,26 @@ Declaration.Scoped finishRecord(String anonName) {
...
-            List<MemoryLayout> prevBitfieldLayouts = bitfieldLayouts;
             List<Declaration> prevBitfieldDecls = bitfieldDecls;
+            long prevBitfieldSize = bitfieldSize;
             bitfieldDecls = null;
+            bitfieldSize = 0;
             if (!prevBitfieldDecls.isEmpty()) {
-                addField(bitfield(prevBitfieldLayouts, prevBitfieldDecls.toArray(new Declaration.Variable[0])));
+                addField(offset, bitfield(prevBitfieldDecls.toArray(new Declaration.Variable[0])));
+            }
+            if (prevBitfieldSize == 0) {
```
mcimadamore, line 158: This `if` is new - there is a change in semantics from Java 20 to Java 21 - as Java 20 disallows creation of zero-length padding, while Java 21 allows it (which seems like a bug). So, to make jextract work against Java 20, this change was required.

JornVernee (approved):

PR #121, opened 2023-05-30
7903481: Jextract doesn't enforce group layout alignment correctly in some cases

mcimadamore:
This patch overhauls the treatment of pragma packs directives.

The current logic tries to detect fields occurring at misaligned offsets, and relaxes alignment constraints for these fields.
However, in cases like this:

```
#pragma pack(push, 1)
struct A {
   long long a;
   int b;
}
```

Each field is correctly aligned. But the struct size (12) is not a multiple of its natural alignment (8). As a result, we run into issues when building a sequence layout out of this struct, because of the eager checks added to the layout API.

This patch fixes support for packed structs "the right way", that is, by asking clang the struct/union alignment, and then making sure that any field is aligned accordingly before the group layout is created.

---------
### Progress

### Issue
 * [CODETOOLS-7903481](https://bugs.openjdk.org/browse/CODETOOLS-7903481): Jextract doesn't enforce group layout alignment correctly in some cases

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/121/head:pull/121` \
`$ git checkout pull/121`

`$ git checkout pull/121` \
`$ git pull https://git.openjdk.org/jextract.git pull/121/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 121`

`$ git pr show -t 121`

</details>

<a href="https://git.openjdk.org/jextract/pull/121.diff">https://git.openjdk.org/jextract/pull/121.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/121#issuecomment-1568306011)

### DIFF
--- a/src/main/java/org/openjdk/jextract/impl/TypeImpl.java
+++ b/src/main/java/org/openjdk/jextract/impl/TypeImpl.java
@@ -373,7 +373,7 @@ public abstract class TypeImpl implements Type {
     public static Optional<MemoryLayout> getLayout(org.openjdk.jextract.Type t) {
         try {
             return Optional.of(getLayoutInternal(t));
-        } catch (Throwable ex) {
+        } catch (UnsupportedOperationException ex) {
             return Optional.empty();
         }
     }
@@ -389,7 +389,7 @@ public abstract class TypeImpl implements Type {
             } else {
                 return Optional.of(FunctionDescriptor.of(getLayoutInternal(retType), args));
             }
-        } catch (Throwable ex) {
+        } catch (UnsupportedOperationException ex) {
             return Optional.empty();
         }
     }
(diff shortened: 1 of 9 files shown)


src/main/java/org/openjdk/jextract/impl/TypeImpl.java
```diff
@@ -373,7 +373,7 @@ public String toString() {
     public static Optional<MemoryLayout> getLayout(org.openjdk.jextract.Type t) {
         try {
             return Optional.of(getLayoutInternal(t));
-        } catch (Throwable ex) {
+        } catch (UnsupportedOperationException ex) {
```
mcimadamore, line 376: These changes are not strictly necessary, but I don't like that we swallow important exceptions and return empty optionals instead.

mcimadamore:
Note, even with this fix, windows.h does not extract correctly with the latest `panama` branch. This seems to be caused by a bug in the logic with which we visit record fields - either because of a bug in jextract, or because of an issue in libclang itself. Essentially, it is possible for jextract to see struct fields in the "wrong order", which then completely messes up our offset computation logic. More investigation is required to fix that case (esp. to understand if that's a latent issue in jextract code).

JornVernee (approved):

PR #137, opened 2023-10-22
7903574: jextract should add lang=c to snippets containing C sources

wangweij:
add the tag when snippet is generated.

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Change must be properly reviewed (no review required)

### Issue
 * [CODETOOLS-7903574](https://bugs.openjdk.org/browse/CODETOOLS-7903574): jextract should add lang=c to snippets containing C sources (**Bug** - P4)

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - Committer)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/137/head:pull/137` \
`$ git checkout pull/137`

Update a local copy of the PR: \
`$ git checkout pull/137` \
`$ git pull https://git.openjdk.org/jextract.git pull/137/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 137`

View PR using the GUI difftool: \
`$ git pr show -t 137`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jextract/pull/137.diff">https://git.openjdk.org/jextract/pull/137.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/137#issuecomment-1774083481)

PR #139, opened 2023-11-20
Update jextract 21 with latest fixes

mcimadamore:
This PR brings three changesets to the `master` branch, so that we can release updated binaries:

* https://git.openjdk.org/jextract/pull/132
* https://git.openjdk.org/jextract/pull/131
* https://github.com/openjdk/jextract/pull/137

### DIFF
--- a/src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
+++ b/src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
@@ -213,7 +213,7 @@ abstract class ClassSourceBuilder extends JavaSourceBuilder {
             append("\n");
         }
         indent();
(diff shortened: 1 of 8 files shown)


src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
```diff
@@ -213,7 +213,7 @@ void emitDocComment(Declaration decl, String header) {
             append("\n");
         }
         indent();
-        append(" * {@snippet :\n");
+        append(" * {@snippet lang=c :\n");
```
mcimadamore, line 216: This is a new fix - see: https://github.com/openjdk/jextract/pull/140

JornVernee (approved):

PR #141, opened 2023-11-22
7903590: Refactor jextract source code generation

JornVernee:
We currently use `ClassSourceBuilder` both for top level classes and nested classes. That means for instance that we need to check whether to generate package declarations and import statements, and for nested classes we also emit code into a separate intermediate buffer that is copied into the enclosing buffer when we end the nested class. This PR proposes splitting part of the logic in `ClassSourceBuilder` into a separate `SourceFileBuilder`, which handles package decl/imports, and holds the underling `StringBuilder` which allows a nested class to share the `StringBuilder` with its enclosing class. Each `SourceFileBuilder` produces exactly one `JavaFileObject`. `TopLevelBuilder` and `Constants` now aggregate over `SourceFileBuilder`s instead of `JavaSourceBuilder`s. This also allows reducing the number of classes that implement `JavaSourceBuilder` to just `TopLevelBuilder` and `StructBuilder`, and allows dropping several methods from the interface. Other builders now wrap `SourceFileBuilder`, and emit code into its buffer.

Jextract source code generation related code currently relies on method overriding to allow sub classes to inject behavior into the generation code in a super class. This, however, makes it hard to follow what method implementation is actually being called when reading code. And, any method call may become 'suspicious' because it might do something non-obvious. For example, `emitConstructor` is overridden to do nothing in several sub-classes. But, this is not obvious when seeing the `emitConstructor()` call in `ClassSourceBuilder::begin`.

Furthermore, we use method overriding for instance to specify the class modifiers for a `ClassSourceBuilder` subclass, which means that control flow will flow through several class levels when generating code. This, and other cases, could be replaced by a simple (`String`) value passed to `ClassSourceBuilder` constructor. Using a direct field value in the generation code makes it clear that there is nothing special going on (as the method call might imply).

This PR proposes to remove most uses of method overriding, to make it easier to follow control flow when reading the code. This has a few implications:
-  Direct field access is used where possible, and data is passed to *Builder constructors instead of through a virtual method call to a sub class implementation
- `emitConstructor` was not needed in all cases. So, I've moved it from `ClassSourceBuilder::classBegin` to callers of that method instead. This allows the call to be omitted in some cases, without the need to define a sub class that overrides `emitConstructor` to do nothing. (Also renamed to `emitPrivateDefaultConstructor` for clarity).
- Doc comments for methods need to be indented. This was previously handled by overriding the `emitDocComment` method in `HeaderFileBuilder` to add the indentation. Instead, I've moved the call to `emitDocComment` to the code generation `emit*` methods, which now handle the indentation for both the doc comment and the code together.
- The `TopLevelBuilder.SplitHeader` and `FirstHeader` sub classes are no longer needed. They can be replaced by `HeaderFileBuilder`. I've create a factory method that takes over part of the code gen logic that `FirstHeader` was injecting by overriding `classBegin`.
- Sub classes that need to inject extra code gen, now have their own generation methods which compose calls to `classBegin`/`classEnd` with other code gen methods. `classBegin` and `classEnd` now do the dumb/obvious thing. A good example of this is the new `generate` method in `FunctionalInterfaceBuilder`:

```
    void generate() {
        emitDocComment(funcType, className());
        classBegin();
        emitFunctionalInterfaceMethod();
        emitFunctionalFactories();
        emitFunctionalFactoryForPointer();
        classEnd();
    }
```

I quite like how all the code here is in one place.


src/main/java/org/openjdk/jextract/impl/Constants.java
```diff
@@ -71,14 +73,20 @@ public Constants(JavaSourceBuilder enclosing) {
...
-        if (currentBuilder.constantIndex > CONSTANTS_PER_CLASS || currentBuilder == null) {
-            if (currentBuilder != null) {
-                currentBuilder.classEnd();
-            }
-            currentBuilder = new Builder(currentBuilder.enclosing, constantBuilders.size());
-            constantBuilders.add(currentBuilder);
-            currentBuilder.classBegin();
+        if (currentBuilder.constantIndex > CONSTANTS_PER_CLASS) {
+            currentBuilder.classEnd();
+            currentBuilder = startNewBuilder(packageName, constantBuilders.size());
```
JornVernee, line 89: `currentBuilder` is never `null`, so we don't need to check for that.

src/main/java/org/openjdk/jextract/impl/Constants.java
```diff
@@ -245,10 +260,6 @@ final class NamedConstant extends Constant {
                 this.constantName = newConstantName();
             }
 
-            String constantName() {
-                return constantName;
-            }
-
```
JornVernee, line 251: This method was unused, so I've removed it.

src/main/java/org/openjdk/jextract/impl/Constants.java
```diff
@@ -213,26 +238,16 @@ static Constant ofLiteral(Class<?> type, Object value) {
...
 
-        Builder(JavaSourceBuilder encl, int id) {
-            super(encl, Kind.CLASS, "constants$" + id);
-        }
+        private static final String MEMBER_MODS = "static final";
 
-        String memberMods() {
-            return kind == ClassSourceBuilder.Kind.CLASS ?
-                    "static final " : "";
-        }
```
JornVernee, line 231: Note that `Builder`s always have the `CLASS` kind, so we can just use a `static final String` for the modifiers.

src/main/java/org/openjdk/jextract/impl/OutputFactory.java
```diff
@@ -179,17 +179,18 @@ public Void visitScoped(Declaration.Scoped d, Declaration parent) {
         }
 
         boolean isStructKind = Utils.isStructOrUnion(d);
+        JavaSourceBuilder prevBuilder = null;
```
JornVernee, line 182: Note that `StructBuilder` no longer keeps a reference to the enclosing builder. So, it can not be returned from `classEnd`/`end`. Instead, we keep track of the enclosing builder here.

This also helps prevent StructBuilder from being too context aware.

src/main/java/org/openjdk/jextract/impl/SourceFileBuilder.java
```diff
@@ -0,0 +1,124 @@
...
+        align--;
+    }
+
+    int align() {
+        return align;
+    }
+
+    public JavaFileObject toFile(Function<String, String> finisher) {
+        return Utils.fileFromString(packageName, className, finisher.apply(sb.toString()));
+    }
```
JornVernee, line 119: Used by `TopLevelBuilder::toFiles` to replace the placeholder `#{SUPER}` super class name for the first header.

src/main/java/org/openjdk/jextract/impl/StructBuilder.java
```diff
@@ -81,60 +88,71 @@ private List<String> prefixNamesList() {
...
-    void classDeclBegin() {
+    void begin() {
         if (!inAnonymousNested()) {
+            if (isNested()) {
+                sourceFileBuilder().incrAlign();
+            }
             emitDocComment(structTree);
+            classBegin();
+            Constant layoutConstant = constants.addLayout(((Type.Declared) structType).tree().layout().orElseThrow());
+            layoutConstant.emitGetter(this, MEMBER_MODS, Constant::nameSuffix);
```
JornVernee, line 103: Note that I didn't get rid of _all_ the context dependent logic in this patch. That would require more surgery in OutputFactory, which I've left alone for now.

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
```diff
@@ -50,160 +47,104 @@ enum Kind {
...
-        String res = build();
-        sb = null;
-        return List.of(Utils.fileFromString(packageName(), className(), res));
     }
 
     // Internal generation helpers (used by other builders)
 
-    void append(Object o) {
-        sb.append(o);
-    }
```
JornVernee, line 165: This overload, and the one accepting a `boolean` were unused, so I've dropped them.

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
```diff
@@ -50,160 +47,104 @@ enum Kind {
...
 
-    boolean isNested() {
-        return enclosing instanceof ClassSourceBuilder;
+    final String className() {
+        // for a (nested) class 'com.foo.package.A.B.C' this will return 'C'
+        return className;
     }
 
-    String className() {
-        return desc.displayName();
```
JornVernee, line 75: Note that `desc.displayName()` is inaccurate for nested classes, as it will return e.g. `A$B` for a nested `A.B` class. This is then impossible to distinguish from class names that actually have a `$` in the name. I felt that it was better to just use a `String` to represent the simple class name instead.

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
```diff
@@ -235,42 +176,11 @@ void emitDocComment(Type.Function funcType, String name) {
         append(" */\n");
     }
 
-    // is the name enclosed enclosed by a class of the same name?
-    boolean isEnclosedBySameName(String name) {
-        return className().equals(name) ||
-                (isNested() && enclosing.isEnclosedBySameName(name));
-    }
```
JornVernee, line 242: Moved to `StructBuilder`, which is the only user of this method.

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
```diff
@@ -235,42 +176,11 @@ void emitDocComment(Type.Function funcType, String name) {
...
-            append("import java.lang.invoke.MethodHandle;\n");
-            append("import java.lang.invoke.VarHandle;\n");
-            append("import java.nio.ByteOrder;\n");
-            append("import java.lang.foreign.*;\n");
-            append("import static java.lang.foreign.ValueLayout.*;\n");
-        }
-    }
-
-    void emitConstantGetter(String mods, String getterName, boolean nullCheck, String symbolName, Constant constant) {
+    final void emitConstantGetter(String mods, String getterName, boolean nullCheck, String symbolName, Constant constant, Declaration decl) {
```
JornVernee, line 179: This method is only used by `Constants,Builder` so maybe it should be moved there. I've left it for now.

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
```diff
@@ -50,160 +47,104 @@ enum Kind {
...
-    // code buffer
-    private StringBuilder sb = new StringBuilder();
-    // current line alignment (number of 4-spaces)
-    private int align;
-
-    ClassSourceBuilder(JavaSourceBuilder enclosing, Kind kind, String name) {
-        this.enclosing = enclosing;
-        this.align = (enclosing instanceof ClassSourceBuilder classSourceBuilder)
-                ? classSourceBuilder.align : 0;
+    ClassSourceBuilder(SourceFileBuilder builder, String modifiers, Kind kind, String className, String superName, List<String> enclosingClassNames) {
```
mcimadamore, line 57: If this models a class, and classes can be nested, it is not clear to me as to why we wouldn't have a pointer to the enclosing class, rather than having just a list of names?

src/main/java/org/openjdk/jextract/impl/SourceFileBuilder.java
```diff
@@ -0,0 +1,124 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package org.openjdk.jextract.impl;
+
+import javax.tools.JavaFileObject;
+import java.lang.constant.ClassDesc;
+import java.util.function.Function;
+
+public class SourceFileBuilder {
```
mcimadamore, line 31: Should this class be final?

src/main/java/org/openjdk/jextract/impl/TypedefBuilder.java
```diff
@@ -27,29 +27,21 @@
...
-    }
-
-    @Override
-    String superClass() {
-        return superClass;
     }
 
-    @Override
-    void classDeclBegin() {
+    void generate() {
```
mcimadamore, line 41: This method seems only to be called after we create a new `TypedefBuilder`. Same is true for `FunctionalInterfaceBuilder`. It's also not a special method, just something that clients need to remember to call.
I'm wondering if this patch isn't really trying to transition away from the "building" metaphore - e.g. what you want is to create a new "typedef", "functional interface" object, and add it to some list of generated objects (so that their code will be all co-generated at the end). Having the client to call `generate` explicitly so that the string buffer is filled seems redundant. Either the string buffer is filled at construction (eagerly), or it is filled when creating the java file object (lazily).

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
JornVernee: Yeah, that might be simpler. I started out trying to remove the reference to enclosing classes altogether, but that turned out to be impossible since we might need to reference a nested class from source code we generate.

src/main/java/org/openjdk/jextract/impl/TypedefBuilder.java
mcimadamore: Not suggesting to remove those builders. But I think the client saying `new FunctionalInterfaceBuilder(...)` should be enough.

src/main/java/org/openjdk/jextract/impl/SourceFileBuilder.java
JornVernee: Yes. Also, it doesn't need to be `public`. The same goes for `FunctionalInterfaceBuilder` and `TypedefBuilder`

src/main/java/org/openjdk/jextract/impl/TypedefBuilder.java
JornVernee: Okay. In that case I'd like to go with the static `generate` method. In terms of contract surface that is the same, but it allows giving a name to the code generation 'side effect' as well.

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
```diff
@@ -48,19 +48,19 @@ enum Kind {
...
 
-    ClassSourceBuilder(SourceFileBuilder builder, String modifiers, Kind kind, String className, String superName, List<String> enclosingClassNames) {
+    ClassSourceBuilder(SourceFileBuilder builder, String modifiers, Kind kind, String className, String superName, ClassSourceBuilder enclosing) {
         this.sb = builder;
         this.modifiers = modifiers;
         this.kind = kind;
         this.className = className;
         this.superName = superName;
-        this.enclosingClassNames = enclosingClassNames;
+        this.enclosing = enclosing;
```
mcimadamore, line 63: should we check that the enclosing class has the same source file builder as this one?

mcimadamore (approved):

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
JornVernee: Note really important from the `ClassSourceBuilder` perspective. Having builders with different underlying `SourceFileBuilders` would still work, the sources would just be emitted to different source files.

PR #145, opened 2023-11-29
7903586: Revisit jextract constant classes

mcimadamore:
This PR removes the use of jextract constant classes, and falls back to a simpler code generation strategies which move the various constants closer to where they belong:

* for structs and callbacks, the constants are added to the class/interface being generated;
* for the header class, we use a local holder class per function to hold the method handle and function descriptor constants.

This scheme results in much less generated code (as no constant class is emitted), but with the same performance and laziness as the old scheme.

When working with the code, I've decided to simplify the handling of constant generation, in the spirit of what was recently done with the introduction of string templates. That is, instead of using a magic `Constant` class, which has methods to generate getters/setters for a given constant, I converted much of the code into "static" string templates, so that it's easier to see which code gets generated when staring at the jextract code. As a result, the `Constant` class has been removed.

---------
### Progress

### Issue
 * [CODETOOLS-7903586](https://bugs.openjdk.org/browse/CODETOOLS-7903586): Revisit jextract constant classes (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/145/head:pull/145` \
`$ git checkout pull/145`

`$ git checkout pull/145` \
`$ git pull https://git.openjdk.org/jextract.git pull/145/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 145`

`$ git pr show -t 145`

</details>

<a href="https://git.openjdk.org/jextract/pull/145.diff">https://git.openjdk.org/jextract/pull/145.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/145#issuecomment-1831690888)


src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
```diff
@@ -171,16 +180,118 @@ final void emitDocComment(Type.Function funcType, String name) {
             """);
     }
 
-    final void emitConstantGetter(String mods, String getterName, boolean nullCheck, String symbolName, Constant constant, Declaration decl) {
-        incrAlign();
-        if (decl != null) {
-            emitDocComment(decl);
+    public String mangleName(String javaName, Class<?> type) {
```
mcimadamore, line 183: most of the methods added here come from the old `Constant` class

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
```diff
@@ -171,16 +180,118 @@ final void emitDocComment(Type.Function funcType, String name) {
...
-                return \{nullCheck ? STR."RuntimeHelper.requireNonNull(\{constant}, \"\{symbolName}\")" : constant};
+    }
+
+    String layoutString(int textBoxIndent, MemoryLayout l) {
+        StringBuilder builder = new StringBuilder();
+        layoutString(textBoxIndent, builder, l);
+        return builder.toString();
+    }
+
+    private void layoutString(int textBoxIndent, StringBuilder builder, MemoryLayout l) {
```
mcimadamore, line 209: Note: the logic is a bit convoluted because we have to add some indentation to the resulting string, as it can span multiple lines. And so that it is correctly aligned inside the text box in which it appears into.

mcimadamore: Ideally, we could avoid all this by just using low-level append calls - but that would make it hard for clients to just embed the layout inside a string template, and make the code less readable, so I went for this instead.

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
```diff
@@ -155,7 +164,7 @@ final void emitDocComment(Declaration decl, String header) {
             /**
             \{!header.isEmpty() ? STR." * \{header}\n" : ""}\
              * {@snippet lang=c :
-             \{CDeclarationPrinter.declaration(decl, " ".repeat(align()*4) + " * ")}
+            \{CDeclarationPrinter.declaration(decl, " * ")}
```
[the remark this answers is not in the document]
mcimadamore: This also brought in some changes in `CDeclarationPrinter` which was adding newlines on its own, which now conflicted with string template usage (e.g I could see extra lines being inserted in the javadoc).

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -241,4 +230,110 @@ private void emitGlobalSetter(Constant segmentConstant, Constant vhConstant, Str
             """);
         decrAlign();
     }
+
+    public String emitGlobalSegment(MemoryLayout layout, String javaName, String nativeName, Declaration declaration) {
```
mcimadamore, line 234: This is a recurrent theme: sometimes the "emit" method will return the name of a relevant variable that has been generated, so that other code can refer to it.

src/main/java/org/openjdk/jextract/impl/FunctionalInterfaceBuilder.java
```diff
@@ -38,28 +37,29 @@
 
 final class FunctionalInterfaceBuilder extends ClassSourceBuilder {
 
+    private static final String MEMBER_MODS = "static";
```
JornVernee, line 40: Seems unused? (I think I prefer just 'inlining' the mods in the string template if they don't change, as you've done)

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -241,4 +230,110 @@ private void emitGlobalSetter(Constant segmentConstant, Constant vhConstant, Str
...
+    }
+
+    private String emitVarLayout(MemoryLayout layout, String javaName) {
+        String mangledName = mangleName(javaName, MemoryLayout.class);
+        appendIndentedLines(STR."""
+            private static final MemoryLayout \{mangledName} = \{layoutString(0, layout)};
+
+            \{MEMBER_MODS} MemoryLayout \{mangledName}() {
+                return \{mangledName};
+            }
```
JornVernee, line 260: Now that we no longer have the constants classes, I wonder if we should just do away with the getter here and just expose the layout field.

I noticed that we are also already exposing the FunctionDescriptor in a functional interface class as a field, as part of this patch,

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -241,4 +230,116 @@ private void emitGlobalSetter(Constant segmentConstant, Constant vhConstant, Str
             """);
         decrAlign();
     }
+
+    public String emitGlobalSegment(MemoryLayout layout, String javaName, String nativeName, Declaration declaration) {
+        String mangledName = mangleName(javaName, MemorySegment.class);
+        incrAlign();
+        appendLines(STR."""
+            private static final MemorySegment \{mangledName} = RuntimeHelper.lookupGlobalVariable(\"\{nativeName}\", \{layoutString(0, layout)});
```
JornVernee, line 238: Shouldn't this re-use the layout field we emit instead of emitting the same layout string again inline?

JornVernee (approved):

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
mcimadamore: I wondered that also. I think that, for the context of this PR, I'd like to avoid changes to the shape of the generated code. Should be easy to fix as a followup change. As observed offline, some getters perform extra validation (e.g. null checks).

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
mcimadamore: good point

JornVernee (approved):

PR #148, opened 2023-11-30
Simplify jextract backend pipeline

mcimadamore:
Now that we have mutable attributes, we can make the jextract pipeline a little simpler, by making sure that each step in the jextract pipeline doesn't destroy the information contained in the jextract model and, additionally, by making sure that each step can only add attributes to the IR (which are acted upon later on).

To do this, I had to do some foundational changes to TreeMaker, to make sure that all declaration nodes were being correctly de-duplicated: since declarations are referenced back by types, if a declaration is re-created multiple times, the types pointing to that declaration end up being out of sync, and that's not desirable (because that could lead to a loss of some of the attributes).

I also removed the `Transformer` interface: altering the model is not something the jextract pipeline should attempt to do. This meant that existing transformers have been reimplemented as side-effecting visitors, as follows:

* `EnumConstantLifter` adds an attribute to constants which belong to an enum (but no longer drops the enum on the ground);
* `IncludeFilter` marks several declaration with a `Skip` attribute, to signal `outputFactory` not to generate any code for them;
* `DuplicateFilter` also uses the `Skip` attribute to drop redundant declarations.

Additionally, the `NameMangler` has been simplified quite a bit, to the point that now it is just a simple visitor which ends up attaching attributes to declarations (`JavaName`, `JavaParameterNames`, `JavaFunctionalInterfaceName`). These attributes are later inspected (from `OutputFactory`). Since now the attribute is added directly to the relevant declaration, there is no need to have complex keying logic inside `NameMangler`.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/148/head:pull/148` \
`$ git checkout pull/148`

`$ git checkout pull/148` \
`$ git pull https://git.openjdk.org/jextract.git pull/148/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 148`

`$ git pr show -t 148`

</details>

<a href="https://git.openjdk.org/jextract/pull/148.diff">https://git.openjdk.org/jextract/pull/148.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/148#issuecomment-1834413623)


src/main/java/org/openjdk/jextract/impl/IncludeFilter.java
```diff
@@ -76,43 +69,35 @@ public Void visitScoped(Declaration.Scoped d, Declaration parent) {
...
-        decls.add(tree);
         return null;
     }
 
     @Override
     public Void visitVariable(Declaration.Variable tree, Declaration parent) {
         if (parent == null && !includeHelper.isIncluded(tree)) {
-            return null;
+            //skip
+            tree.addAttribute(new Skip());
```
JornVernee, line 94: Maybe we could use a singleton instance?

src/main/java/org/openjdk/jextract/impl/NameMangler.java
```diff
@@ -213,25 +144,27 @@ public Void visitFunction(Declaration.Function func, Declaration parent) {
...
     }
 
     @Override
     public Void visitScoped(Declaration.Scoped scoped, Declaration parent) {
-        String name = scoped.name().isEmpty()? parent.name() : scoped.name();
-        if (declJavaNames.containsKey(new NameAndDecl(name, scoped))) {
-            //skip struct that's seen already
-            return null;
+        String name = scoped.name();
+        if (name.isEmpty() && parent != null) {
```
JornVernee, line 163: Can we see a `null` parent here now? The old code doesn't have the null check.

src/main/java/org/openjdk/jextract/impl/TreeMaker.java
```diff
@@ -103,18 +103,25 @@ public Declaration createTree(Cursor c) {
...
+                case FunctionDecl -> createFunction(c);
+                case StructDecl -> createRecord(c, Declaration.Scoped.Kind.STRUCT);
+                case UnionDecl -> createRecord(c, Declaration.Scoped.Kind.UNION);
+                case TypedefDecl -> createTypedef(c);
+                case VarDecl -> createVar(c, Declaration.Variable.Kind.GLOBAL);
+                default -> null; // skip
+            };
+            declarationCache.put(pos, decl);
+        }
+        return decl;
```
JornVernee, line 124: IIRC we had a cache here before, but removed it because it was faster to re-create trees. That might have inadvertently also created latent issues with inconsistent IR views of the same decl?

Any way, we need the cache to 'canonicalize' IR nodes. Good to know.

src/main/java/org/openjdk/jextract/impl/IncludeFilter.java
mcimadamore: Yeah...

src/main/java/org/openjdk/jextract/impl/TreeMaker.java
mcimadamore: Yes, we used to have a cache and removed it - I don't think we thought too much about what it meant in terms of losing indentity of the objects involved, since we never relied on that anyway. But I remember this being a problem when we took our first stab at the name mangling patch, which was confirmed with an experiment I did few days ago: w/o de-duplication, it is possible to set an attribute on a struct, but to still have types in the IR that point to the old, un-attributes struct, which is bad.

src/main/java/org/openjdk/jextract/impl/DeclarationImpl.java
```diff
@@ -345,13 +341,91 @@ public int hashCode() {
         }
     }
 
+    // attributes
```
mcimadamore, line 344: Now all attributes, and relevant wither/getter/predicates are all defined here.

JornVernee: Nice!

PR #149, opened 2023-12-01
7903585: Revisit variadic support

JornVernee:
Currently, we only support var args passed to a variadic function through an `Object[]`. From that, we infer the layouts of the arguments, and then link a variadic method handle on the fly.

This is problematic because: 1. there's an ambiguity between structs and pointers, since they both use `MemorySegment` as a carrier. 2. from a performance perspective, re-linking the method handle every time is somewhat costly.

The approach proposed by this patch is to add a variadic invoker factory method, which can, given a list of layouts, be used to create an instance of a variadic invoker (interface), that embeds a downcall handle specialized to a particular type of the variadic function. The user specifies the layout explicitly, so there is no ambiguity (1), and the user can cache the returned invoker and re-use it instead of re-linking the function on every invocation (2).

The current layout inferring method can then be re-implemented on top of this factory.

Question: I'm currently generating the doc comment on both the inferring method and the invoker factory. Is this good? Should it be on one or the other? Should it be on the invoker interface as well?

---

There were some issues I ran into while developing this patch, that I fixed:
- We had some methods in the tests that take no arguments, but were declared without a prototype, leading to them getting translated as variadic functions in the bindings. I've change those parameter lists to `(void)` (this popped out when I only had the invoker factory in an earlier iteration)
- I noticed that jtreg was sometimes not recompiling a test class in the generator tests after the bindings changed, leading to weird NoSuchMethodErrors. I've fixed this by adding an explicit `@build` step to the tests.
- I've added `-retain:fail,error` to the jtreg arguments, so that the generated sources for a failing test are not deleted immediately, and can be manually inspected after a test failure.
- I noticed we weren't generating a layout for `C_BOOL`, because we don't treat that type as being supported. I think this is a latent issue. I've added Bool as a supported type to fix this.

Also: I've ret-conned Test8244959 and renamed it to TestPrintf. And added new cases for the invoker.

---------
### Progress

### Issue
 * [CODETOOLS-7903585](https://bugs.openjdk.org/browse/CODETOOLS-7903585): Revisit variadic support (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/149/head:pull/149` \
`$ git checkout pull/149`

`$ git checkout pull/149` \
`$ git pull https://git.openjdk.org/jextract.git pull/149/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 149`

`$ git pr show -t 149`

</details>

<a href="https://git.openjdk.org/jextract/pull/149.diff">https://git.openjdk.org/jextract/pull/149.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/149#issuecomment-1836158131)


test/jtreg/generator/funcPointerInvokers/TestFuncPointerInvokers.java
```diff
@@ -36,12 +36,14 @@
...
   * @library /lib
  * @run main/othervm JtregJextract -l Func -t test.jextract.funcpointers func.h
+ * @build TestFuncPointerInvokers
  * @run testng/othervm --enable-native-access=ALL-UNNAMED TestFuncPointerInvokers
  */
 /*
  * @test id=sources
   * @library /lib
- * @run main/othervm JtregJextract -l Func -t test.jextract.funcpointers func.h
+ * @run main/othervm JtregJextractSources -l Func -t test.jextract.funcpointers func.h
```
JornVernee, line 45: Note that here we were invoking the wrong binding generator class.

src/main/resources/org/openjdk/jextract/impl/resources/RuntimeHelper.java.template
```diff
@@ -99,147 +110,60 @@ final class RuntimeHelper {
...
-            if (needsAllocator) {
-                mh = mh.bindTo(allocator);
-            }
-            // flatten argument list so that it can be passed to an asSpreader MH
-            Object[] allArgs = new Object[nNamedArgs + unnamedArgs.length];
-            System.arraycopy(args, 0, allArgs, 0, nNamedArgs);
-            System.arraycopy(unnamedArgs, 0, allArgs, nNamedArgs, unnamedArgs.length);
-
-            return mh.asSpreader(Object[].class, argsCount).invoke(allArgs);
+    private static Class<?> unboxIfNeeded(Class<?> clazz) {
```
mcimadamore, line 113: question - wouldn't it be simpler to have a single switch from carrier type to variadic layout directly? I think the current code organization is the way it is because we used to need both carriers and function descriptor, but now we only need the latter.

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -170,6 +168,44 @@ class Holder {
                 }
             }
             """);
+        } else {
+            String invokerName = javaName + "$invoker";
+            String invokerFactoryName = javaName + "$makeInvoker";
+            String paramExprs = paramExprs(declType, finalParamNames, isVarArg);
+            appendLines(STR."""
+                public interface \{invokerName} {
```
mcimadamore, line 176: question: if we used a record here, could we perhaps get rid of the holder class (since record fields are trusted) ? We don't expect this interface to be created by the user, so there's maybe not much value in it being a functional interface?

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -170,6 +168,44 @@ class Holder {
...
+                    \{retType} \{javaName}(\{paramExprs});
+                }
+
+                """);
+            emitDocComment(decl);
+            appendLines(STR."""
+                public static \{invokerName} \{invokerFactoryName}(MemoryLayout... layouts) {
+                    class Holder {
+                        static final FunctionDescriptor BASE_DESC = \{descriptorString(2, descriptor)};
```
mcimadamore, line 185: Why is the descriptor inside the holder class but the MH is not?

mcimadamore: Ugh - because the method handle depends on the layout parameter. But if that's the case, what's the benefit of the holder class? I'd thought that to get maximum performance, the specialized method handle had to be a constant.

JornVernee: We re-link the MH every time, so it is not a constant. Note that it depends on `layouts`.

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
JornVernee: Not sure what you mean. The holder class is for the base descriptor. We could stick that into the invoker interface instead, but I'm not sure why a record would be better than an interface? Do we want users to be able to access the underlying MethodHandle?

test/jtreg/generator/testPrintf/TestPrintf.java
```diff
@@ -0,0 +1,123 @@
...
+ */
+/*
+ * @test id=sources
+ *
+ * @library /lib
+ * @run main/othervm JtregJextractSources -t test.jextract.printf -l Printf printf.h
+ * @build TestPrintf
+ * @run testng/othervm --enable-native-access=ALL-UNNAMED TestPrintf
+ */
+public class TestPrintf {
```
mcimadamore, line 51: Nice test!

mcimadamore (approved):

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
mcimadamore: Sorry, I guess what I'm trying to say is that:
* the benefit for caching the base descriptor is low;
* the real benefit would be in treating the final (specialized) method handle as a constant

To do the latter, given we can't use stable fields, the only way to get there is to use a record. So, if we care about maximum performance, we should use a record. If we don't care, an interface is fine. In both cases, I don't think there's much to gain in having an extra holder class here.

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
mcimadamore: Yes, but if a client wants to stash an invoker with given shape in a static final field, then being able to trust that MH would be important I think.

JornVernee: Field of lambda classes are trusted right? So the captured MH will be a trusted instance field?

src/main/resources/org/openjdk/jextract/impl/resources/RuntimeHelper.java.template
JornVernee: Thanks for the tip. I've managed to simplify this code quite a bit: https://github.com/openjdk/jextract/pull/149/commits/a8c387c89a4ea00dabbbc50479994c5e6b4714a4

- We only care about wrapper types, not primitves, since all the arguments come to us in an `Object[]`.
- Smaller-than-int types should be promoted to `int`, not `long`.
- No need to handle `void` (in fact, that would be wrong).

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
JornVernee: FWIW, I had a record in an earlier iteration, but realized it wasn't needed, and using a lambda expression makes the generated code a bit more concise.

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
JornVernee: Okay, I can drop the holder class.

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
JornVernee: Ok, I did a quick experiment, and the JIT doesn't seem to be able to constant fold through a captured MH field. But it does work for a record. I will switch to using a record then

src/main/resources/org/openjdk/jextract/impl/resources/RuntimeHelper.java.template
```diff
@@ -110,60 +110,18 @@ final class RuntimeHelper {
...
-        }
-        if (MemorySegment.class.isAssignableFrom(c)) {
-            return MemorySegment.class;
-        }
-        throw new IllegalArgumentException("Invalid type for ABI: " + c.getTypeName());
-    }
-
     private static MemoryLayout variadicLayout(Class<?> c) {
-        if (c == long.class) {
+        // apply default argument promotions per C spec
```
mcimadamore, line 114: Wow, that's much simpler :-)

mcimadamore (approved):

PR #150, opened 2023-12-01
7903597: Slim down RuntimeHelper

JornVernee:
Remove some of the helper methods in RuntimeHelper. They can be moved inline in the generated code.

Also: I've elected to move the `System.load(Library)` directives we generated inside RuntimeHelper into the first header class, in order to have it more localized to the actual code that a user sees. That also makes RuntimeHelper more 'dumb'. Only the package declaration is now injected into the template, which makes it easier for users to combined multiple extraction runs (they can just discard one of the generated RuntimeHelpers).

---------
### Progress

### Issue
 * [CODETOOLS-7903597](https://bugs.openjdk.org/browse/CODETOOLS-7903597): Slim down RuntimeHelper (**Enhancement** - P4)

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - Committer) ⚠️ Review applies to [b87c2a5d](https://git.openjdk.org/jextract/pull/150/files/b87c2a5dac4704413cb60f151e624b60b5a62331)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/150/head:pull/150` \
`$ git checkout pull/150`

`$ git checkout pull/150` \
`$ git pull https://git.openjdk.org/jextract.git pull/150/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 150`

`$ git pr show -t 150`

</details>

<a href="https://git.openjdk.org/jextract/pull/150.diff">https://git.openjdk.org/jextract/pull/150.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/150#issuecomment-1836557449)


test/jtreg/generator/testLinkageErrors/TestLinkageErrors.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Test
+    public void nullChecksTest() {
+        assertThrowsULE(() -> func(), "func");
+        assertThrowsULE(() -> func$MH(), "func");
+        assertThrowsULE(() -> x$get(), "x");
+        assertThrowsULE(() -> x$set(1), "x");
+        assertThrowsULE(() -> x$SEGMENT(), "x");
+        assertThrowsULE(() -> y$SEGMENT(), "y");
+        assertThrowsULE(() -> pt$SEGMENT(), "pt");
```
JornVernee, line 70: I've cleaned up this test code a bit, since I had to change the check to look at the exception's cause as well, as we are now throwing the linkage error from the holder class, it can be wrapped into an ExceptionInInitializerError.

src/main/resources/org/openjdk/jextract/impl/resources/RuntimeHelper.java.template
```diff
@@ -90,24 +64,11 @@ final class RuntimeHelper {
     static MethodHandle upcallHandle(Class<?> fi, String name, FunctionDescriptor fdesc) {
         try {
             return MH_LOOKUP.findVirtual(fi, name, fdesc.toMethodType());
-        } catch (Throwable ex) {
+        } catch (ReflectiveOperationException ex) {
```
JornVernee, line 67: I've left this `upcallHandle` method since we use it to initialize a class-level constant in the FI classes. It's useful to have the try/catch be out-of-line like this so we don't have to emit `static` blocks.

src/main/resources/org/openjdk/jextract/impl/resources/RuntimeHelper.java.template
```diff
@@ -2,81 +2,38 @@
...
     static final AddressLayout POINTER = ValueLayout.ADDRESS.withTargetLayout(MemoryLayout.sequenceLayout(MAX_VALUE, JAVA_BYTE));
 
-    final static SegmentAllocator CONSTANT_ALLOCATOR =
-            (size, align) -> Arena.ofAuto().allocate(size, align);
-
     static {
-        #LOAD_LIBRARIES#
         SymbolLookup loaderLookup = SymbolLookup.loaderLookup();
-        SYMBOL_LOOKUP = name -> loaderLookup.find(name).or(() -> LINKER.defaultLookup().find(name));
+        Linker linker = Linker.nativeLinker();
```
mcimadamore, line 27: Question: with RuntimeHelper being so small now, and given one of the stated goal is to allow for better composition of different extractions, would it not be convenient to move all these definitions in the header class and ditch `RuntimeHelper` completely? If we do that, we might also avoid generating the varargs definitions unless we need them, so that the header class would only include minimal additional code.

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -272,16 +291,16 @@ private void emitGlobalSetter(String segmentConstant, String vhConstant, String
...
-            private static final MemorySegment \{mangledName} = RuntimeHelper.lookupGlobalVariable(\"\{nativeName}\", \{layout});
-
-            """);
         if (declaration != null) {
             emitDocComment(declaration);
         }
         appendLines(STR."""
             \{MEMBER_MODS} MemorySegment \{mangledName}() {
-                return RuntimeHelper.requireNonNull(\{mangledName}, \"\{javaName}\");
+                class Holder {
```
mcimadamore, line 299: I was ambivalent on this - but it's probably better to make it consistent with the strategy we use for functions, as you have done.

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -186,7 +189,11 @@ public interface \{invokerName} {
             appendLines(STR."""
                 public static \{invokerName} \{invokerFactoryName}(MemoryLayout... layouts) {
                     FunctionDescriptor baseDesc$ = \{descriptorString(2, descriptor)};
-                    var mh$ = RuntimeHelper.downcallHandleVariadic("\{nativeName}", baseDesc$, layouts);
+                    var mh$ = Linker.nativeLinker().downcallHandle(
```
mcimadamore, line 192: If we keep other handy generation wrappers, we might keep this as well. It's not super trivial after all.

src/main/resources/org/openjdk/jextract/impl/resources/RuntimeHelper.java.template
mcimadamore: I don't disagree. In fact I think I'm in favor of also leaving downcall handle (in both variants).

mcimadamore (approved):
Looks very good. I think I'm mildly leaning for moving the common code in the header class (which is out true main point), and perhaps make some of the code generation conditional. Note also that, now that we only have few methods to add, we could use string templates/text blocks rather than a template file, which will simplify things a bit (no more reading files etc.)

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -238,6 +245,18 @@ void emitPointerTypedef(Declaration.Typedef typedefTree, String name) {
         emitPrimitiveTypedefLayout(name, TypeImpl.PointerImpl.POINTER_LAYOUT, typedefTree);
     }
 
+    void emitLoadLibraries(List<String> libraries) {
```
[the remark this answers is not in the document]
JornVernee: Right, I was thinking of dropping it, but I've kept it for now, since we also fall back to the default lookup, and I'm not sure how important that is.

src/main/resources/org/openjdk/jextract/impl/resources/RuntimeHelper.java.template
JornVernee: I thought about this, and I think the main question is: in which header class do we generate the helpers if we have multiple classes? I wasn't sure if we'd be happy with duplicating the helpers in every header class (users might complain about the duplication). But, maybe the majority of cases will only ever see a single header class, so there's still a benefit.

I'll take another look at that. The conditional code generation is also an interesting benefit.

JornVernee: For fields that is not possible though, since we have to put those at the start of the header class to avoid illegal forward references, but we don't know when we create the first header class whether we are going to need multiple header classes. When we have only 1 header class, the first header is the 'root' into which the helpers should go, when we have multiple headers, the second header class is the root.

I think maybe it's simpler to just generate the helper methods and fields like `SYMBOL_LOOKUP` once per header class. Or use static methods with embedded Holder classes for the fields. We could also have a special last header that everything else inherits, i.e. just have the last header extend RuntimeHelper essentially. But when I got to that point in my thought process, I thought: why not just keep RuntimeHelper then.

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
```diff
@@ -58,14 +58,17 @@ enum Kind {
     private final String className;
     private final String superName;
     private final ClassSourceBuilder enclosing;
+    private final String runtimeHelperName;
 
-    ClassSourceBuilder(SourceFileBuilder builder, String modifiers, Kind kind, String className, String superName, ClassSourceBuilder enclosing) {
+    ClassSourceBuilder(SourceFileBuilder builder, String modifiers, Kind kind, String className, String superName,
+                       ClassSourceBuilder enclosing, String runtimeHelperName) {
```
mcimadamore, line 64: nit: maybe we want to call this "mainHeaderClass" or something like that?

mcimadamore (approved):
Looks very good, I think dropping RuntimeHelper will help developers better understanding what comes out of an extraction run

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
JornVernee: I had `firstHeaderName` at first, but I didn't like how that name doesn't signal what the purpose of the field is. So, I switched it to `runtimeHelperName` (there might be a better name). If we ever decide to change the location of the helper methods again, the name would still work, since the purpose stays the same.

PR #152, opened 2023-12-04
7903600: Add more tests for unsupported types

JornVernee:
Add more tests for unsupported types. See the new TestUnsupportedTypes.java

While working on this, I also noticed several cases of dead code: `FunctionalInterfaceScanner` is completely unused, and `InMemoryJavaCompiler::jfoFromByteArray` and `InMemoryJavaCompiler::jfoFromString` (and its caller in `OutputFactory`) are unused as well. I've removed them in this PR.

Changes in this PR:
- I'm passing in the error stream PrintWriter that we create in the test all the way down to UnsupportedFilter, so that it can write to it instead of `System.err`. Then we can check what was written in the test. 
- I've added warning print outs for the different reasons why we are skipping declarations. Some of these were missing.
- I've removed a check for ValueLayouts that are larger than 8 bytes in `visitVariable`. If the type of a variable is an unsupported primitive, we already filter it out earlier on in the method, so there's no need to check again.
- I've added a missing `Skip.with` in `visitVariable`, so that we will filter out (global) variables with unsupported function types.
- I've added skipping of typedefs when they have an unsupported type or a type that does not have a layout. For the latter, I've also had to remove a test case that checked if we generated a typedef class for an undefined struct type. Since the struct is undefined, I believe we should not generate a typedef binding class for that case.

The new test covers all the unsupported type cases that are handled by TestUnsupportedTypes.

---------
### Progress

### Issue
 * [CODETOOLS-7903600](https://bugs.openjdk.org/browse/CODETOOLS-7903600): Add more tests for unsupported types (**Enhancement** - P4)

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - Committer) ⚠️ Review applies to [cb930413](https://git.openjdk.org/jextract/pull/152/files/cb9304133702dece2f589c853be598c111eba331)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/152/head:pull/152` \
`$ git checkout pull/152`

`$ git checkout pull/152` \
`$ git pull https://git.openjdk.org/jextract.git pull/152/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 152`

`$ git pr show -t 152`

</details>

<a href="https://git.openjdk.org/jextract/pull/152.diff">https://git.openjdk.org/jextract/pull/152.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/152#issuecomment-1839457489)


test/lib/testlib/JextractToolRunner.java
```diff
@@ -146,7 +146,6 @@ protected static JextractResult run(Object... options) {
     protected static JextractResult run(String... options) {
         StringWriter writer = new StringWriter();
         PrintWriter pw = new PrintWriter(writer);
-        String[] args = new String[options.length + 1];
```
JornVernee, line 149: Also dead code.

test/testng/org/openjdk/jextract/test/toolprovider/Test8245767.java
```diff
@@ -44,10 +42,6 @@ public void testTypedefs() {
             Class<?> cls = loader.loadClass("test8245767_h");
             assertNotNull(cls);
 
-            // class should be generated for typedef on opaque struct
-            Class<?> fooCls = loader.loadClass("Foo");
-            assertNotNull(fooCls);
-
```
JornVernee, line 50: This is the problematic test case that I believe to be testing for behavior that we do not want.

src/main/java/org/openjdk/jextract/impl/UnsupportedFilter.java
```diff
@@ -161,20 +186,21 @@ public Void visitDeclaration(Declaration d, Declaration declaration) {
         return null;
     }
 
-    private boolean checkFunctionTypeSupported(Declaration decl, Type.Function func) {
+    private boolean checkFunctionTypeSupported(Declaration decl, Type.Function func, String nameOfSkipped) {
         String unsupportedType = firstUnsupportedType(func);
         if (unsupportedType != null) {
-            warn("skipping " + JavaName.getOrThrow(decl) + " because of unsupported type usage: " +
-                    unsupportedType);
+            warnUnsupportedType(nameOfSkipped, unsupportedType);
```
mcimadamore, line 192: Should we consolidate this with the other warnings? At the end of the day, we're skipping here because of an unsupported type, so I'd prefer to see `warnSkip` being used, where "unsupported type" appears in the "cause"

src/main/java/org/openjdk/jextract/impl/UnsupportedFilter.java
```diff
@@ -95,31 +103,32 @@ public Void visitFunction(Function funcTree, Declaration parent) {
...
-            warn("skipping " + name + " because of unsupported type usage: " +
-                    unsupportedType);
+            warnUnsupportedType(name, unsupportedType);
             Skip.with(varTree);
             return null;
         }
         MemoryLayout layout = Type.layoutFor(varTree.type()).orElse(null);
         if (layout == null) {
             //no layout - skip
+            warnSkip(name, "can not compute MemoryLayout");
```
mcimadamore, line 116: I suggest "does not have a valid memory layout/descriptor"

test/testng/org/openjdk/jextract/test/toolprovider/Test8245767.java
mcimadamore: I've seen that yesterday, and was equally puzzled by the "expected behavior"

test/testng/org/openjdk/jextract/test/toolprovider/unsupported/TestUnsupportedTypes.java
```diff
@@ -0,0 +1,74 @@
...
+package org.openjdk.jextract.test.toolprovider.unsupported;
+
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+import testlib.JextractToolRunner;
+
+import java.nio.file.Path;
+
+public class TestUnsupportedTypes extends JextractToolRunner {
```
mcimadamore, line 32: Nice and comprehensive test - good to have everything in one place!

mcimadamore (approved):

src/main/java/org/openjdk/jextract/impl/UnsupportedFilter.java
JornVernee: I'll do this. It introduces a bit of duplication between the warning messages, but it's probably clearer for the reader.

PR #153, opened 2023-12-04
Add build support for generating Jacoco coverage reports

JornVernee:
Add a new `coverage` gradle task that can be used to generate a Jacoco test coverage report. Because we're on the bleeding edge, the user running the build needs to build Jacoco from source for this to work. Though, this is optional.

As an example: I've cloned Jacoco from here: https://github.com/jacoco/jacoco and built the project using:

    mvn package -DskipTests -Dbytecode.version=22
    
Then, I point gradle at the generated artifacts using:

    jacoco_cli=<jacoco repo>/org.jacoco.cli/target/org.jacoco.cli-0.8.12-SNAPSHOT-nodeps.jar
    jacoco_agent=<jacoco repo>/org.jacoco.agent/target/classes/jacocoagent.jar
    
I've put these in a local `gradle.properties` file instead of passing them on the command line.

I then run `./gradlew coverage` which generates an html coverage report under `./build/jacoco-report`.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/153/head:pull/153` \
`$ git checkout pull/153`

`$ git checkout pull/153` \
`$ git pull https://git.openjdk.org/jextract.git pull/153/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 153`

`$ git pr show -t 153`

</details>

<a href="https://git.openjdk.org/jextract/pull/153.diff">https://git.openjdk.org/jextract/pull/153.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/153#issuecomment-1839555859)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -190,27 +190,64 @@ task cmakeBuild(type: Exec) {
     ]
 }
 
-// run jtreg tests. Note: needs jtreg_home variable set to point to the jtreg
-task jtreg(type: JavaExec) {
-    dependsOn createRuntimeImageForTest,cmakeBuild
+void createJtregTask(String name, boolean coverage, String os_lib_dir) {
+    tasks.register(name, JavaExec) {
+        dependsOn createRuntimeImageForTest,cmakeBuild
+
+        doFirst {
+            if (findProperty("jtreg_home") == null) {
+                throw new GradleException("jtreg_home is not defined")
+            }
+            // e.g.: <jacoco repo>/org.jacoco.agent/target/classes/jacocoagent.jar
+            if (coverage && findProperty("jacoco_agent") == null) {
+                throw new GradleException("jacoco_agent is not defined")
+            }
+        }
+
+        workingDir = "$buildDir"
+
+        classpath = files(findProperty("jtreg_home") + "/lib/jtreg.jar")
+
+        args = [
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -190,27 +190,60 @@ task cmakeBuild(type: Exec) {
...
+
+        args += "../test"
     }
+}
 
-    workingDir = "$buildDir"
+createJtregTask("jtreg", false, os_lib_dir)
+createJtregTask("jtregWithCoverage", true, os_lib_dir)
+
+tasks.register("coverage", JavaExec) {
```
JornVernee, line 232: Note that I'm using `tasks.register` here to make the configuration of this task lazy. That allows us to avoid an error when the `jacoco_cli` is not set, but the task isn't invoked.

The old jtreg task did this in a `doFirst` block, but that code actually doesn't run until the task itself is invoked. So, if there is a missing property, and the configuration references that property, we'll get an NPE instead of the intended exception.

Note that I've moved the property null check out of the `doFirst` block of the jtreg task as well, and made the configuration lazy there too. If I do that, I see an exception as expected when the task is configured, but not when the task doesn't run.

mcimadamore (approved):

PR #154, opened 2023-12-06
7903602: Simplify handling of structs/unions 

mcimadamore:
This PR simplifies how declarations and layouts for C structs/unions are created, by splitting the monolithic `RecordLayoutComputer` into two passes:

1. As part of parsing (in `TreeMaker`) a scoped declaration is created for a struct, which contains all the relevant fields. This declaration is annotated with a bunch of attributes which reify some of the properties of the underlying clang cursors/types (such as sizes, alignments and offsets).
2. We now have a new static method in `Declaration`, namely `Declaration::layoutFor` which allows clients to obtain the layout for a given declaration. For most declaration, this method will get the type and then call `Type::layoutFor`. But for scoped declarations we will compute a new struct/union layout.
3. The already computed layouts are attached to the scoped declaration using an attribute, so that we avoid computing things multiple times.

I actually started off with a declaration visitor which attached layouts to all scoped declarations, but then reverted to a more lazy approach like the one described in (2) because this approach is more amenable for tests dynamically querying layouts of parsed declarations (whereas if layouts depended on a certain visitor being run after parsing, all such tests could not be supported).

Note that the logic which computes a struct/union layout is designed so that it always yields the same answer. That is, even if a client asks the layout for a nested anon struct first, the layout is computed correctly. This is possible because I've removed a lot of stateful logic from the computation, namely:

* the name of anonymous nested declaration is uniquely derived from the declaration position
* the starting offset used for layout computation is derived from the offset of the first member of the scoped declaration (if we're seeing an anonymous struct, otherwise the starting offset is just zero)

An interesting consequence of this PR is that now the Declaration/Type API become layout-agnostic: while it's possible to compute a layout/function descriptor from a declaration or type, the declaration/type IRs do not embed any layout information. This change is also evident in the treatment for primitive types, which no longer have a pre-determined associated layout. That is, jextract declaration are now pure wrappers around clang cursors, and nothing more. Of course given the information in a declaration we can derive a layout, but that's an orthogonal operation: a declaration exists even w/o a layout.

Finally, I've cleaned up some scoped kinds (class, namespace) - and corresponding factories - which were not really used.

---------
### Progress

### Issue
 * [CODETOOLS-7903602](https://bugs.openjdk.org/browse/CODETOOLS-7903602): Simplify handling of structs/unions (**Bug** - P3)

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to [1dc1b6a1](https://git.openjdk.org/jextract/pull/154/files/1dc1b6a197262ea487dae73a22b3778419e07dbd)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/154/head:pull/154` \
`$ git checkout pull/154`

`$ git checkout pull/154` \
`$ git pull https://git.openjdk.org/jextract.git pull/154/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 154`

`$ git pr show -t 154`

</details>

<a href="https://git.openjdk.org/jextract/pull/154.diff">https://git.openjdk.org/jextract/pull/154.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/154#issuecomment-1843055695)


src/main/java/org/openjdk/jextract/impl/TreeMaker.java
```diff
@@ -234,18 +236,73 @@ public Declaration.Scoped createRecord(Cursor c, Declaration.Scoped.Kind scopeKi
...
-                allDecls.add(createTree(child));
+                decls.add(createTree(child));
             }
         });
-        List<Declaration> decls = filterNestedDeclarations(allDecls);
         if (c.isDefinition()) {
             //just a declaration AND definition, we have a layout
-            MemoryLayout layout = TypeMaker.valueLayoutForSize(c.type().size() * 8).layout().orElseThrow();
-            return Declaration.enum_(CursorPosition.of(c), c.spelling(), layout, decls.toArray(new Declaration[0]));
+            return Declaration.enum_(CursorPosition.of(c), c.spelling(), decls.toArray(new Declaration[0]));
```
JornVernee, line 305: `TypeMaker.valueLayoutForSize` is now unused, so it can be removed.

src/main/java/org/openjdk/jextract/impl/TreeMaker.java
```diff
@@ -234,18 +236,73 @@ public Declaration.Scoped createRecord(Cursor c, Declaration.Scoped.Kind scopeKi
...
+                } else {
+                    if (!pendingBitFields.isEmpty()) {
+                        pendingFields.add(Declaration.bitfields(pendingBitfieldsPos.get(), pendingBitFields.toArray(Variable[]::new)));
+                        pendingBitFields.clear();
+                        pendingBitfieldsPos.set(null);
+                    }
+                    if (fc.isAnonymousStruct()) {
+                        // process struct recursively
+                        pendingFields.add(recordDeclaration(parent, fc).tree());
+                    } else if (!fc.isBitField() && !fc.spelling().isEmpty()) {
```
JornVernee, line 264: The condition here seems to be unnecessary? We should already be in the no-bitfield case due to enclosing if/else, and the isAnonymousStruct check should filter out fields without a name?

JornVernee (approved):

src/main/java/org/openjdk/jextract/impl/MacroParserImpl.java
```diff
@@ -337,7 +337,7 @@ void updateTable(TypeMaker typeMaker, Cursor decl) {
 
         void reparseMacros(boolean recovery) {
             String snippet = macroDecl(recovery);
-            TreeMaker treeMaker = new TreeMaker();
+            TreeMaker treeMaker = new TreeMaker(); // @@@: what about de-duplicated declarations?
```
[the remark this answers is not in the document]
mcimadamore: The comment is meant to indicate that, since we're creating a brand new TreeMaker, all the types we create in here, esp. those with a declared type, will point to declaration that are NOT deduplicated. I have a followup fix for this.

mcimadamore:
As discussed offline, I will follow up on this patch with a change to make the size/alignment/offset information part of the jextract Type/Declaration APIs, rather than having them be as optional attributes, which, I think, is closer to the spirit of what the declaration API wants to achieve.

PR #155, opened 2023-12-08
7903605: Simplify TypeMaker

mcimadamore:
The main goal of this PR is to make TypeMaker simpler, and remove the convoluted logic which stashes pointers, and asks client to call a method to resolve their types.

Instead, this patch tweaks the behavior of the TypeMaker so that when a pointer to a declared type is created, it points to a special supplier. This supplier, when called, looks in the declaration cache for a matching declaration (in the `TreeMaker`). If a corresponding declaration is found, a new declared type pointing to that declaration is returned as a pointee type.

It is also possible for a pointer to point to an opaque struct, in which case there might be no symbol in the declaration cache. In such case we make the supplier return an erroneous type instead.

The patch also addresses a number of related issues:

* Since declarations are stable identity-wise but types are not, I removed the `Attributed` interface and corresponding implementation class. Now all the methods to deal with attributes are in `Declaration` - declarations are the only entities that can be reliably attributed (because of de-duplication).

* I added a mechanism to create "nested" `TreeMaker` - that is, a `TreeMaker` that can delegate a symbol lookup to a parent `TreeMaker`. This is handy when dealing with macros, where we do not want to pollute the original environment, but we still want to be able to resolve declarations included in that environment.

* I have fixed the equals methods in Declaration - these methods ignored attributes and were not working correctly. As a result, I also did some changes in duplicate filter, as using declaration equality does not work (because attributes might mismatch thanks to the added `Skip`).

---------
### Progress

### Issue
 * [CODETOOLS-7903605](https://bugs.openjdk.org/browse/CODETOOLS-7903605): Simplify TypeMaker (**Enhancement** - P3)

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to [92b29ef7](https://git.openjdk.org/jextract/pull/155/files/92b29ef71d5a7ad4bab91251f0227d59080993a7)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/155/head:pull/155` \
`$ git checkout pull/155`

`$ git checkout pull/155` \
`$ git pull https://git.openjdk.org/jextract.git pull/155/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 155`

`$ git pr show -t 155`

</details>

<a href="https://git.openjdk.org/jextract/pull/155.diff">https://git.openjdk.org/jextract/pull/155.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/155#issuecomment-1847912873)


src/main/java/org/openjdk/jextract/impl/TreeMaker.java
```diff
@@ -111,20 +124,23 @@ public Declaration createTree(Cursor c) {
...
-                default -> null; // skip
-            };
+        Optional<Declaration> cachedDecl = lookup(pos);
+        if (cachedDecl.isPresent()) {
+            return cachedDecl.get();
+        }
+        Declaration decl = switch (c.kind()) {
+            case EnumDecl -> createEnum(c);
+            case EnumConstantDecl -> createEnumConstant(c);
+            case FieldDecl -> createVar(c, Declaration.Variable.Kind.FIELD);
```
JornVernee, line 134: FWIW, I see the `FieldDecl` case as being never executed in the code coverage report. Not sure why...

src/main/java/org/openjdk/jextract/impl/TreeMaker.java
```diff
@@ -261,8 +278,8 @@ final Type.Declared recordDeclaration(Cursor parent, Cursor recordCursor) {
                     if (fc.isAnonymousStruct()) {
                         // process struct recursively
                         pendingFields.add(recordDeclaration(parent, fc).tree());
-                    } else {
-                        Type fieldType = typeMaker.makeType(fc.type());
+                    } else if (!fc.isBitField() && !fc.spelling().isEmpty()) {
```
JornVernee, line 281: Spurious change? Or was this condition needed after all?

src/main/java/org/openjdk/jextract/impl/TreeMaker.java
```diff
@@ -343,7 +360,7 @@ private Declaration.Typedef createTypedef(Cursor c) {
             Type pointeeType = null;
             try {
                 pointeeType = ((Type.Delegated)canonicalType).type();
-            } catch (NullPointerException npe) {
+            } catch (IllegalStateException unresolved) {
```
JornVernee, line 363: This catch block is never taken according to coverage report. Maybe it can just be removed?

JornVernee (approved):

src/main/java/org/openjdk/jextract/impl/TreeMaker.java
mcimadamore: This (and other changes) have come back as a result of a bad merge I will fix

src/main/java/org/openjdk/jextract/impl/TreeMaker.java
mcimadamore: I think this is caused by the fact that the new `TreeMaker::recordDeclaration` (introduced in previous PR) does all the processing of a struct/union inline.

src/main/java/org/openjdk/jextract/impl/TreeMaker.java
```diff
@@ -111,20 +124,23 @@ public Declaration createTree(Cursor c) {
...
+            case FieldDecl -> createVar(c, Declaration.Variable.Kind.FIELD);
+            case ParmDecl -> createVar(c, Declaration.Variable.Kind.PARAMETER);
+            case FunctionDecl -> createFunction(c);
+            case StructDecl -> createRecord(c, Declaration.Scoped.Kind.STRUCT);
+            case UnionDecl -> createRecord(c, Declaration.Scoped.Kind.UNION);
+            case TypedefDecl -> createTypedef(c);
+            case VarDecl -> createVar(c, Declaration.Variable.Kind.GLOBAL);
+            default -> null; // skip
+        };
+        if (decl != null && pos != Position.NO_POSITION) {
```
mcimadamore, line 143: If position is NO_POSITION, we do not cache. But I wonder, if there's no position (which indicates some clang intrinsics like va_list), should we really even generate a declaration?

mcimadamore: There is a difference in the two approaches suggested above. Consider this example:

```
#include <stdarg.h>

struct Foo {
    int x;
    va_list v;
};
```

If we simply skip all declarations with NO_POSITION, we get a weird layout for Foo:

```
private static final MemoryLayout $LAYOUT = MemoryLayout.structLayout(
        JAVA_INT.withName("x"),
        MemoryLayout.paddingLayout(4),
        MemoryLayout.paddingLayout(24)
    ).withName("Foo");
```

Whereas with this patch we get this:

```
private static final MemoryLayout $LAYOUT = MemoryLayout.structLayout(
        JAVA_INT.withName("x"),
        MemoryLayout.paddingLayout(4),
        MemoryLayout.sequenceLayout(1,         MemoryLayout.structLayout(
                JAVA_INT.withName("gp_offset"),
                JAVA_INT.withName("fp_offset"),
                foo_h.C_POINTER.withName("overflow_arg_area"),
                foo_h.C_POINTER.withName("reg_save_area")
            ).withName("__va_list_tag")    ).withName("v")
    ).withName("Foo");
```

Both layouts are "correct". The struct interface won't have any accessor for the `v` field no matter which approach is taken (which makes sense, as we have no real declaration for `va_list`). While the second layout is more precise, I'm starting to think that the first approach is more honest: if we don't really have a cursor position for a declaration, and we cannot deduplicate, we cannot guarantee that declaration attributes won't be "lost" (as they might be added to different copies of the same underlying intrinsic cursor). WDYT?

JornVernee: I think it may ultimately be preferable to keep as much of the layout intact, even if we don't generate accessors. That way, the user could do something with the layout themselves if they wanted to. I think maybe deduplicating based on `Position` only is not enough in the presence of intrinsics which do not have a position. But, we can have another look at better intrinsic support later as well. I think it's fine to just drop them for now.

PR #159, opened 2023-12-13
7903609: Jextract can not handle some anonymous nested structs

JornVernee:
Improve the handling of nested anonymous records (structs and unions).

The current way in which we find the offset of a nested anonymous record is by getting the offset of the first field relative to the first named (or unnamed outermost) parent. However, if the first field doesn't have a name, we bail out, as getting the offset of an unnamed field is not supported by libclang.

But, we can reliably compute the offset of a nested anonymous record that has at least one named field nested somewhere inside of it, by: 1.) getting the offset of the named field to the first named parent record. 2.) getting the offset of the first named field to the anonymous record for which we want to know the offset (also nested inside the same named parent). 3.) computing the offset of the anonymous record by subtracting the offset found in 2. by the offset found in 1.

This is also sufficient to address cases like these:

     struct Foo {
         char c; // offset = 0
         struct <anon1> { // offset = 96 - 64 = 32
             int: 32;
             struct <anon2> { // offset = 96 - 32 = 64
                 int: 32;
                 int x; // offset(Foo) = 96, offset(anon2) = 32, offset(anon1) = 64
             };
         };
     };

This offset is attached through the `AnonymousStruct` attribute during parsing (which is where we have the Cursors needed to compute these offsets).

Furthermore, this PR contains some additional improvements:
- We were manually skipping `BITFIELDS` scoped declaration in the record layout generating code. I've changed this to use a `Skip` that is added to the BITFIELDS declaration, rather than to the individual bitfields inside of it. We can then just look for the `Skip` when generating the layout, and that also handles skipped nested anonymous records correctly (which don't have any named fields).
- I also noticed that we don't print out the name of a skipped bitfield correctly, because we use the name of the immediate parent to generated the name of the field, which for a bitfield is always empty, since the immediate parent is a scoped BITFIELDS decl. This resulted in warnings about e.g. `.a` being skipped. I'm now propagating the first named parent instead of the immediate parent, which can then be used to generate a valid field name.

---------
### Progress

### Issue
 * [CODETOOLS-7903609](https://bugs.openjdk.org/browse/CODETOOLS-7903609): Jextract can not handle some anonymous nested structs (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/159/head:pull/159` \
`$ git checkout pull/159`

`$ git checkout pull/159` \
`$ git pull https://git.openjdk.org/jextract.git pull/159/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 159`

`$ git pr show -t 159`

</details>

<a href="https://git.openjdk.org/jextract/pull/159.diff">https://git.openjdk.org/jextract/pull/159.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/159#issuecomment-1854272943)

### DIFF
--- a/src/main/java/org/openjdk/jextract/impl/TreeMaker.java
+++ b/src/main/java/org/openjdk/jextract/impl/TreeMaker.java
@@ -350,8 +350,12 @@ class TreeMaker {
                     result.set(OptionalLong.of(offsetToOutermost - offsetToAnon));
                     return false;
                 } else if (fc.isAnonymousStruct()) {
-                    result.set(offsetOfAnonymousRecord(outermostParent, anonRecord, fc));
-                    return false;
+                    OptionalLong nestedResult = offsetOfAnonymousRecord(outermostParent, anonRecord, fc);
+                    if (nestedResult.isPresent()) {
+                        result.set(nestedResult);
+                        return false;
+                    }
+                    return true;
                 }
             }
             return true;
--- a/src/main/java/org/openjdk/jextract/impl/TreeMaker.java
+++ b/src/main/java/org/openjdk/jextract/impl/TreeMaker.java
@@ -33,6 +33,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.OptionalLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 
@@ -274,7 +275,8 @@ class TreeMaker {
                     Type fieldType = toType(fc);
                     Variable bitfieldDecl = Declaration.bitfield(CursorPosition.of(fc), fc.spelling(), fc.getBitFieldWidth(), fieldType);
                     if (!fc.spelling().isEmpty()) {
-                        ClangOffsetOf.with(bitfieldDecl, parent.type().getOffsetOf(fc.spelling()));
+                        long offset = parent.type().getOffsetOf(fc.spelling());
+                        ClangOffsetOf.with(bitfieldDecl, offset);
                     }
                     pendingBitFields.add(bitfieldDecl);
                 } else {
@@ -290,7 +292,8 @@ class TreeMaker {
                         Declaration fieldDecl = createTree(fc);
                         ClangSizeOf.with(fieldDecl, fc.type().kind() == TypeKind.IncompleteArray ?
                                 0 : fc.type().size() * 8);
-                        ClangOffsetOf.with(fieldDecl, parent.type().getOffsetOf(fc.spelling()));
+                        long offset = parent.type().getOffsetOf(fc.spelling());
+                        ClangOffsetOf.with(fieldDecl, offset);
                         ClangAlignOf.with(fieldDecl, fc.type().align() * 8);
                         pendingFields.add(fieldDecl);
                     }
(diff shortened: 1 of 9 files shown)


src/main/java/org/openjdk/jextract/impl/TreeMaker.java
```diff
@@ -315,12 +316,47 @@ final Type.Declared recordDeclaration(Cursor parent, Cursor recordCursor) {
...
+     *     struct <anon1> { // offset = 96 - 64 = 32
+     *         int: 32;
+     *         struct <anon2> { // offset = 96 - 32 = 64
+     *             int: 32;
+     *             int x; // offset(Foo) = 96, offset(anon2) = 32, offset(anon1) = 64
+     *         };
+     *     };
+     * };
+     */
+    private static OptionalLong offsetOfAnonymousRecord(Cursor outermostParent, Cursor anonRecord, Cursor record) {
```
mcimadamore, line 343: What about DeclarationImpl::recordMemberOffset ? Is that still used? UnsupportedTypeFilter and StructBuilder depend on that for skipping - is is still the case that the first offset might be missing?

JornVernee: yes, it's still used. The offset of an anon struct might be missing if it doesn't have any named fields nested inside. In that case, the offset is OptionalLong.empty(), and the UnsupportedFilter adds a Skip to the record.

src/main/java/org/openjdk/jextract/impl/TreeMaker.java
```diff
@@ -315,12 +316,47 @@ final Type.Declared recordDeclaration(Cursor parent, Cursor recordCursor) {
...
+        AtomicReference<OptionalLong> result = new AtomicReference<>(OptionalLong.empty());
+        record.forEach(fc -> {
+            if (result.get().isPresent()) return;
```
mcimadamore, line 346: Why don't we return after setting a result, instead of waiting for the next cursor and then returning if we see a result has been set?

JornVernee: We can not return from the outer function in the forEach. I don't think there is a way to short-circuit the forEach.

JornVernee: We could enhance the forEach to take a `Predicate` instead, so that we can return true/false to short-circuit, and then have a Consumer overload that just always returns true. I'll look at that

JornVernee: Done in: https://github.com/openjdk/jextract/pull/159/commits/876dd3612b915aced226cce665b8318382445f0e

mcimadamore: Sorry, got mixed up. You just want to avoid further processing if the result has already been set

JornVernee:
I've turned on verbose output for failed jtreg tests: https://github.com/openjdk/jextract/pull/159/commits/363f04a1c9e4dc11c5398a46b980177de5fbd653

So, that I can see why the new test is failing in GHA.

I think this mode is probably good to use in general? (the same as in the JDK).

mcimadamore (approved):

JornVernee:
Thanks for the review. There is an issue with the test on Linux. I'll investigate further.

src/main/java/org/openjdk/jextract/impl/TreeMaker.java
JornVernee: Right... I could simplify the check in UnsupportedFilter to look at the off in the attribute directly, instead of going through `recordMemberOffset`. I think in that case we can also simplify `recordMemberOffset` to return a `long` unconditionally.

JornVernee: Addressed in: https://github.com/openjdk/jextract/pull/159/commits/cd00d68d0f0e08912ed358211a663a0fe6274ccb

`recordMemberOffset` is then just used when computing the layout string, so I've move the method to StructBuilder as well.

mcimadamore:
> I've turned on verbose output for failed jtreg tests: [363f04a](https://github.com/openjdk/jextract/commit/363f04a1c9e4dc11c5398a46b980177de5fbd653)
> 
> So, that I can see why the new test is failing in GHA.
> 
> I think this mode is probably good to use in general? (the same as in the JDK).

seems a good idea

JornVernee:
I realized that there is a small bug in the short-circuiting version of forEach I added. If a nested record doesn't contain a named field, we want to keep looking since there might be another nested record with a named field. Addressed in: https://github.com/openjdk/jextract/pull/159/commits/c44caad14e249f6316c88af0ed8337822ba0313f

mcimadamore (approved):
Good catch!

PR #160, opened 2023-12-14
7903608: Cyclic initialization leads to NPE in header class with global variable

mcimadamore:
This PR fixes some new static initialization cycles that have been made possible by (a) the removal of constant classes and (b) the fact that we refer to struct layouts by name.

The fix consists in two parts:

* First, we make sure that whenever the header class needs to reference another layout, we do so behind a holder class idiom (this patch only touches globals, as all other cases are already using the holder idiom)

* Secondly, we invert the inheritance order of header classes (this is only relevant when using header file splitting). In the current order, symbols that are found "later" in the header files, are added in classes that are "higher up" in the inheritance chain. This leads to issues, as you can have e.g. a primitive typedef depending on a previous primitive typedef, which is in reality defined in a subclass (which leads to a cycle).

Since I was there, I decided to simplify the logic for global variables, and make it more uniform with the rest: now we only emit an accessor for the global segment and a layout. We no longer emit a var handle accessor - developers have all they need to get one. Global vars setters/getters have also been simplified to just use memory segment accessors.

---------
### Progress

### Issue
 * [CODETOOLS-7903608](https://bugs.openjdk.org/browse/CODETOOLS-7903608): Cyclic initialization leads to NPE in header class with global variable (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/160/head:pull/160` \
`$ git checkout pull/160`

`$ git checkout pull/160` \
`$ git pull https://git.openjdk.org/jextract.git pull/160/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 160`

`$ git pr show -t 160`

</details>

<a href="https://git.openjdk.org/jextract/pull/160.diff">https://git.openjdk.org/jextract/pull/160.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/160#issuecomment-1855775267)

### DIFF
--- a/src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
+++ b/src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
@@ -62,10 +62,9 @@ class HeaderFileBuilder extends ClassSourceBuilder {
         if (Utils.isArray(varTree.type()) || Utils.isStructOrUnion(varTree.type())) {
             emitGlobalSegment(layoutVar, javaName, nativeName, varTree);
         } else if (Utils.isPointer(varTree.type()) || Utils.isPrimitive(varTree.type())) {
-            String vhConstant = emitGlobalVarHandle(javaName, layoutVar);
             String segmentConstant = emitGlobalSegment(layoutVar, javaName, nativeName, null);
-            emitGlobalGetter(segmentConstant, vhConstant, javaName, varTree, "Getter for variable:");
-            emitGlobalSetter(segmentConstant, vhConstant, javaName, varTree, "Setter for variable:");
+            emitGlobalGetter(segmentConstant, layoutVar, javaName, varTree, "Getter for variable:");
+            emitGlobalSetter(segmentConstant, layoutVar, javaName, varTree, "Setter for variable:");
 
             if (fiName.isPresent()) {
                 emitFunctionalInterfaceGetter(fiName.get(), javaName);
@@ -293,27 +292,27 @@ class HeaderFileBuilder extends ClassSourceBuilder {
             """);
     }
 
-    private void emitGlobalGetter(String segmentConstant, String vhConstant, String javaName,
+    private void emitGlobalGetter(String segmentConstant, String layoutVar, String javaName,
                                   Declaration.Variable decl, String docHeader) {
         incrAlign();
         emitDocComment(decl, docHeader);
         Class<?> type = Utils.carrierFor(decl.type());
         appendLines(STR."""
             public static \{type.getSimpleName()} \{javaName}$get() {
-                return (\{type.getSimpleName()}) \{vhConstant}.get(\{segmentConstant}(), 0L);
+                return \{segmentConstant}.get(\{layoutVar}, 0L);
             }
             """);
         decrAlign();
     }
 
-    private void emitGlobalSetter(String segmentConstant, String vhConstant, String javaName,
+    private void emitGlobalSetter(String segmentConstant, String layoutVar, String javaName,
                                   Declaration.Variable decl, String docHeader) {
         incrAlign();
         emitDocComment(decl, docHeader);
         Class<?> type = Utils.carrierFor(decl.type());
         appendLines(STR."""
             public static void \{javaName}$set(\{type.getSimpleName()} x) {
-                \{vhConstant}.set(\{segmentConstant}(), 0L, x);
+                \{segmentConstant}.set(\{layoutVar}, 0L, x);
             }
             """);
         decrAlign();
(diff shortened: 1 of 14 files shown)


src/main/java/org/openjdk/jextract/impl/ToplevelBuilder.java
```diff
@@ -77,23 +77,41 @@ private static HeaderFileBuilder createFirstHeader(SourceFileBuilder sfb, List<S
...
-            firstHeader.emitRuntimeHelperMethods();
-        }
         lastHeader.classEnd();
 
         List<JavaFileObject> files = new ArrayList<>();
-        files.add(firstHeader.sourceFileBuilder().toFile(s -> s.replace("extends #{SUPER}",
-                hasOneHeader ? "" : "extends " + lastHeader.className())));
-        files.addAll(builders.stream()
+
+        if (headerBuilders.size() == 1) {
```
mcimadamore, line 89: The changes here are the trickiest. Basically, we can't figure out in advance which name header classes should have. So we end up generating something like:

```
<headerName>_h#{SUFFIX} extends {PREV_SUFFIX}
```

And then leave it to the code here to fixup all the suffixes. For this reason, the name of the runtime helper class cannot just be the current "class name" as that would include the `#{SUFFIX}` bit - but we know that there is going to be _some_ `<headerName>_h` generated at some point, so we can just use that as a runtime helper name.

Overall, I'm not too proud of this code, but at the same time, after evaluating a couple of alternatives they all seemed too complicated. And, we were doing some magic name adaptation even before.

src/main/java/org/openjdk/jextract/impl/Utils.java
```diff
@@ -227,15 +230,19 @@ public static Class<?> carrierFor(Type.Primitive p) {
...
-    public static Class<?> valueLayoutCarrierFor(Type t) {
-        if (t instanceof Delegated delegated && delegated.kind() == Delegated.Kind.POINTER) {
-            return AddressLayout.class;
-        } else if (t instanceof Type.Primitive p) {
-            Class<?> clazz = carrierFor(p);
-            return CARRIERS_TO_LAYOUT_CARRIERS.get(clazz);
-        } else {
-            throw new UnsupportedOperationException(t.toString());
-        }
+    public static Class<?> layoutCarrierFor(Type t) {
```
mcimadamore, line 233: I have generalized this, so that we emit the layout carrier type for any jextract type

src/main/java/org/openjdk/jextract/impl/SourceFileBuilder.java
```diff
@@ -91,8 +91,12 @@ void decrAlign() {
         align--;
     }
 
+    public JavaFileObject toFile(String suffix, Function<String, String> finisher) {
+        return Utils.fileFromString(packageName, STR."\{className}\{suffix}", finisher.apply(sb.toString()));
```
JornVernee, line 95: So, it looks like we adjust the class name at the last moment when generating the file. That's fine I think, but it also means that the `SourceFileBuilder::className` field is 'fake'/a place holder in some cases. That seems like it might be problematic if other code starts to naively use `SourceFileBuilder::className`.

Could we maybe make this more explicit? e.g. could we make the class name an `Optional<String>` ? and then require the name to be injected when calling `toFile`? Or maybe always require the class name to be specified to `toFile`, and just remove the `className` field?

src/main/java/org/openjdk/jextract/impl/ToplevelBuilder.java
JornVernee: > but we know that there is going to be some <headerName>_h generated at some point, so we can just use that as a runtime helper name.

So, just to clarify: we use the last header class we generate as the runtime helper (having no name suffix), but the runtime helper methods and primitive layouts are actually in the first header, and they are visible through static inheritance?

JornVernee (commented):
I think this also fixes: https://bugs.openjdk.org/browse/CODETOOLS-7903607 ?

src/main/java/org/openjdk/jextract/impl/ToplevelBuilder.java
mcimadamore: yes, the last (deepest in the inheritance tree) is used as the runtime helper. That is also the public header class that users depend on. The helper methods are not here, but they are accessible anyway because of inheritance, so it works out.

src/main/java/org/openjdk/jextract/impl/SourceFileBuilder.java
mcimadamore: > So, it looks like we adjust the class name at the last moment when generating the file. That's fine I think, but it also means that the `SourceFileBuilder::className` field is 'fake'/a place holder in some cases. That seems like it might be problematic if other code starts to naively use `SourceFileBuilder::className`.

Yep - I was aware of this, and probably the biggest reason as to why I wasn't proud with the change. I looked at the usages of className the other day, and they felt a bit too many - and depended on in various places when we generate code. I can take another look, but I was in doubt as to whether fix this issue first and maybe improve the code later (given that jextract is currently quite broken). What do you think?

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -367,7 +356,9 @@ private void emitConstant(Class<?> javaType, String constantName, Object value,
             private static final \{javaType.getSimpleName()} \{constantName} = \{constantValue(javaType, value)};
 
             """);
-        emitDocComment(declaration);
+        if (declaration != null) {
+            emitDocComment(declaration);
+        }
```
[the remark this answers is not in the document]
mcimadamore: Good catch - I have absolutely no idea of how that came to be :-)

src/main/java/org/openjdk/jextract/impl/SourceFileBuilder.java
JornVernee: We can address this later as well.

JornVernee (approved):

PR #163, opened 2023-12-14
7903611: Position-based deduplication logic is insufficient

mcimadamore:
This PR adds a new way to deduplicate declarations in TreeMaker that is no longer dependent on cursor positions.
Instead, we now deduplicate using cursor equality (`clang_equalCursors`), which is more precise, and works even when declarations are reused via macros (see test).

The main issue with using cursor equality is that the cursors we want to compare have their segment closed by the time the comparison needs to be performed. So, instead, we dump the contents of the cursor segment on-heap, and stash it in a `Cursor.Key` class. Later, if we need to compare two cursors, we dump the heap contents back on an off-heap segment, and then call `clang_equalCursors` on the two segments. Since we can only compare two cursors at any given time, and we don't use multiple thread, it is safe to share the comparison space.

One problem with using cursor equality is that it doesn't work when the same header is reparsed multiple times - in that case clang will just say that the new cursor and the old cursor are different, even if they point at the same declaration. Presumably that's because the two cursors belong to different translation units.

Because of this, when parsing macros with pointer values, we have to downgrade the pointer type to `void*` (e.g. we lose information on the actual pointee type). This is not too bad, given that this type information was not used anyways.

---------
### Progress

### Issue
 * [CODETOOLS-7903611](https://bugs.openjdk.org/browse/CODETOOLS-7903611): Position-based deduplication logic is insufficient (**Bug** - P3)

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to [c2e4c993](https://git.openjdk.org/jextract/pull/163/files/c2e4c9930f5bd81abab8ee63271b49bd7e40a3aa)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/163/head:pull/163` \
`$ git checkout pull/163`

`$ git checkout pull/163` \
`$ git pull https://git.openjdk.org/jextract.git pull/163/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 163`

`$ git pr show -t 163`

</details>

<a href="https://git.openjdk.org/jextract/pull/163.diff">https://git.openjdk.org/jextract/pull/163.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/163#issuecomment-1857584786)


src/main/java/org/openjdk/jextract/impl/TreeMaker.java
```diff
@@ -63,18 +65,9 @@
...
-    }
+    private final Map<Cursor.Key, Declaration> declarationCache = new HashMap<>();
 
-    private final TreeMaker parent;
-    private final Map<CursorKey, Declaration> declarationCache = new HashMap<>();
-
-    public TreeMaker(TreeMaker parent) {
-        this.parent = parent;
-    }
+    public TreeMaker() { }
```
mcimadamore, line 70: The logic to create "nested" tree makers is gone now - since comparison between cursors from different TUs doesn't work anyways.

test/jtreg/generator/dedup/TestDedup.java
```diff
@@ -0,0 +1,57 @@
+/* Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
```
JornVernee, line 1: It seems that usually there's a blank line at the start of the header

JornVernee (approved):

PR #164, opened 2023-12-15
7903612: Nested struct typedefs fail to compile

mcimadamore:
This PR fixes a small issue with nested struct typedef, and another with typedef in general:

* nested struct classes are emitted with the `final` modifier, which prevents them from being subclassed (e.g. in typedefs)
* all structs have a private constructor to prevent clients from instantiating. Unfortunately this constructor also prevents subclassing (e.g. in typedef).

This patch drops the `final` modifier on nested struct/unions, and also tweaks the default constructor from `private` to `protected`.

### DIFF
--- a/src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
+++ b/src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
@@ -26,7 +26,6 @@ package org.openjdk.jextract.impl;
 
 import org.openjdk.jextract.Declaration;
 import org.openjdk.jextract.Declaration.Constant;
-import org.openjdk.jextract.Declaration.Scoped;
 import org.openjdk.jextract.Type;
 import org.openjdk.jextract.Type.Array;
 import org.openjdk.jextract.Type.Declared;
@@ -35,7 +34,6 @@ import org.openjdk.jextract.Type.Function;
 import org.openjdk.jextract.Type.Primitive;
 import org.openjdk.jextract.impl.DeclarationImpl.JavaName;
 
-import java.lang.foreign.FunctionDescriptor;
 import java.lang.foreign.MemoryLayout;
 import java.lang.foreign.MemorySegment;
 import java.lang.invoke.MethodHandle;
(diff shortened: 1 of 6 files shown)


test/jtreg/generator/packed/TestPackedStructs.java
```diff
@@ -33,7 +33,7 @@
 /*
  * @test id=classes
  * @library /lib
- * @run main/othervm JtregJextract -l Func -t test.jextract.packedstructs packedstructs.h
+ * @run main/othervm JtregJextract -t test.jextract.packedstructs packedstructs.h
```
mcimadamore, line 36: This is an unrelated change: I just noted that this `-l` flag was not needed

JornVernee (approved):
We also don't generate a private constructor for top-level struct classes, but, if we make the constructor `protected` we could generate them there as well. Same for the header classes I think.

If we stick with these constructors, I think it would also be better to make them package-private. That would mean that someone couldn't extend one of these classes from outside of the package (since there's no way to call the super constructor). A practical replacement for the class being `final`.

Alternatively, we could just not have any explicit constructors/final classes. Since all the methods we generate are `static`, there's probably not that much utility to them. Though, I suppose someone might get confused trying to instantiate one of these classes...

mcimadamore:
> We also don't generate a private constructor for top-level struct classes, but, if we make the constructor `protected` we could generate them there as well. Same for the header classes I think.
> 
> If we stick with these constructors, I think it would also be better to make them package-private. That would mean that someone couldn't extend one of these classes from outside of the package (since there's no way to call the super constructor). A practical replacement for the class being `final`.
> 
> Alternatively, we could just not have any explicit constructors/final classes. Since all the methods we generate are `static`, there's probably not that much utility to them. Though, I suppose someone might get confused trying to instantiate one of these classes...

True - package-private is better. I don't think these constructors are necessarily worth it, but for now I just wanted to fix the generated code, we can discuss whether to remove them later.

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
```diff
@@ -137,10 +135,12 @@ void appendIndentedLines(String s) {
         sb.appendIndentedLines(s);
     }
 
-    final void emitPrivateDefaultConstructor() {
-        appendLines(STR."""
-            // Suppresses default constructor, ensuring non-instantiability.
-            private \{className}() {}
+    final void emitDefaultConstructor() {
+        appendIndentedLines(STR."""
```
mcimadamore, line 139: Since I was here, I decided to fix indentation, as I noticed that this is always off - hopefully that's not disruptive with other ongoing work

JornVernee (approved):

PR #167, opened 2023-12-18
7903613: Bad nested names are sometimes attached to structs

mcimadamore:
The `NameMangler` visitor is used to compute the Java name of a jextract declaration. This is implemented as a declaration visitor. Unfortunately, the logic that computes the Java name can be sensitive to the order in which declarations are visited (because this visitor features a "parent" declaration, whose contents affect as to whether a "nested" struct name is generated or not).

In reality, the logic of the name mangler needs to be able to disambiguate between structs that are either anonymous, or already declared somewhere else, and structs that are declared as part of a typedef, variable, function parameter/return declaration. In the former case, we either need no Java name (anonymous struct) or a toplevel Java name. In the latter we need a nested struct name (as the struct class will be nested inside some other class).

This PR introduces a new visitor which tags all struct/union/enum declarations which fall in the latter bucket. This is done with an algorithm which:

1. visits all declarations in a toplevel header
2. remembers which scoped declarations have been seen *directly* (e.g. as part of the visit)
3. keeps track of which scoped declarations can be seen *indirectly* (e.g. because they are behind some declared type)
4. subtracts the declarations in (2) from the declarations in (3), and visits the declarations in the remaining set
5. keeps performing (2), (3), (4) until there's no declaration in (3)

All scoped declarations that appear exclusively as part of some declared type are augmented with the `NestedDecl` attribute, which is then read when calling `Utils::nestedDeclarationFor`. This ensures that all the jextract visitor only recurse on a scoped declaration attached to a type which is known not to have been seen anywhere else. As a result, the behavior of the name mangler is independent of the order in which declarations are seen.

It should be possible, in principle, to leverage this infrastructure to define a declaration visitor that automatically looks inside "nested declarations" (so that subsequent visitors don't really need to concern with following declared types).

I've tested this change with windows.h, which works as expected.

---------
### Progress

### Issues
 * [CODETOOLS-7903613](https://bugs.openjdk.org/browse/CODETOOLS-7903613): Bad nested names are sometimes attached to structs (**Bug** - P3)
 * [CODETOOLS-7903614](https://bugs.openjdk.org/browse/CODETOOLS-7903614): Visit order causes structs to be generated in the wrong enclosing class (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/167/head:pull/167` \
`$ git checkout pull/167`

`$ git checkout pull/167` \
`$ git pull https://git.openjdk.org/jextract.git pull/167/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 167`

`$ git pr show -t 167`

</details>

<a href="https://git.openjdk.org/jextract/pull/167.diff">https://git.openjdk.org/jextract/pull/167.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/167#issuecomment-1861332212)


test/jtreg/generator/nestedInsideAnon/nestedInsideAnon.h
```diff
@@ -0,0 +1,30 @@
...
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+typedef struct {
```
mcimadamore, line 24: This failing test is derived from a failure when extracting windows.h, which was caused by the NestedDecFinder not recursing into a nested decl inside an anonymous struct.

mcimadamore:
Note: this fix is insufficient to fully address issues like this:
https://bugs.openjdk.org/browse/CODETOOLS-7903614

That's because the `NestedDecl` attribute is added to the scoped declaration being declared on the fly - meaning that the same scoped declaration will be reachable from different points in the IR (via different declared types) which might lead jextract to think that the inline declaration is nested in the wrong place.

mcimadamore:
I have updated the fix with a different approach. Instead of relying on a visitor to "guess" what the nested declaration are, I decided to instead attach the declaration corresponding to nested cursors in a new declaration attribute. This means that all the info that clang sees is now reflected in the declaration API. 

Note: libclang is not perfect here, as it often gives jextract duplicate cursors. That is, sometimes the same cursor can appear both at level N and at level N + 1 (this looks like a bug in libclang). The fact that jextract has builtin deduplication for declarations work out in our favor here.

I have tested against `windows.h` again, which extracts fine. I had to make some tweaks so that we only pick up nested "definition" cursors (using `Cursor::isDefinition`), otherwise libclang was sometimes returning weird nested cursors (I saw one for a conditional expression).

src/main/java/org/openjdk/jextract/impl/TreeMaker.java
```diff
@@ -427,7 +426,37 @@ private Declaration.Variable createVar(Cursor c, Declaration.Variable.Kind kind)
...
+        if (!nestedTypes.isEmpty()) {
+            NestedTypes.with(d, nestedTypes);
+        }
+        return d;
+    }
+
+    private void collectNestedTypes(Cursor c, List<Declaration> nestedTypes, boolean ignoreNestedParams) {
+        c.forEach(m -> {
+            if (m.isDefinition()) {
+                if (m.kind() == CursorKind.ParmDecl && !ignoreNestedParams) {
```
mcimadamore, line 453: Note: if we see a param declaration, we need to recursively inspect it, as it might contain more nested declarations. This is useful for function pointer typedefs and variables whose type is a function pointer. In case of a function declaration this logic is disabled, as a function declaration will always point to its parameter declarations anyway.

src/main/java/org/openjdk/jextract/impl/NameMangler.java
```diff
@@ -196,14 +201,45 @@ public Void visitTypedef(Declaration.Typedef typedef, Declaration parent) {
...
         }
+
+        // handle if this typedef is of a struct/union/enum etc.
+        Utils.forEachNested(typedef, d -> {
+            d.accept(this, typedef);
+        });
         return null;
     }
 
+    private String fallbackNameFor(Declaration parent, Declaration.Scoped nested) {
```
mcimadamore, line 212: This method is used to compute a predictable fallback name in case a struct w/o name is found for which a Java name is needed. The general logic is as follows:

* if the parent declaration does not have a function pointer associated, just return the parent name
* otherwise, try to match the nested declaration with a declared type in the function pointer. For arguments declaration we will generate `<parent>$x<num>` whereas for declaration in return position we generate `<parent>$return`.

src/main/java/org/openjdk/jextract/impl/NameMangler.java
```diff
@@ -196,14 +201,45 @@ public Void visitTypedef(Declaration.Typedef typedef, Declaration parent) {
            JavaFunctionalInterfaceName.with(typedef, javaName);
            functionTypeDefNames.put(typedef.type(), javaName);
         }
+
+        // handle if this typedef is of a struct/union/enum etc.
```
mcimadamore, line 205: Note (here and elsewhere): we need to recurse to nested declarations _last_. Otherwise a nested declaration can take the "good" name of the current declaration (which will then be mangled to `<name>$0`).

src/main/java/org/openjdk/jextract/impl/NameMangler.java
```diff
@@ -136,42 +140,46 @@ public Void visitFunction(Declaration.Function func, Declaration parent) {
...
-        String name = scoped.name();
-        if (name.isEmpty() && parent != null) {
-            name = parent.name();
-        }
-        if (JavaName.isPresent(scoped)) {
-            //skip struct that's seen already
-            return null;
-        }
+        if (Utils.isEnum(scoped)) {
+            scoped.members().forEach(fieldTree -> fieldTree.accept(this, null));
```
mcimadamore, line 163: for enums, we don't have to create a new named scope. Also, enums are the only case where jextract leaves an anon enum at the toplevel (where parent == null), so better to skip.

test/testng/org/openjdk/jextract/test/toolprovider/TestClassGeneration.java
```diff
@@ -115,8 +115,8 @@ public static Object[][] structMembers() {
             { "Foo", C_LONG_LONG.withName("ll"), long.class,          10L },
             { "Foo", C_FLOAT.withName("f"),     float.class,         10F },
             { "Foo", C_DOUBLE.withName("d"),    double.class,        10D },
-            { "Bar", C_INT.withName("a"),       int.class,           10 },
-            { "Bar", C_INT.withName("b"),       int.class,           10 },
+            { "Bar$0", C_INT.withName("a"),       int.class,           10 },
```
mcimadamore, line 118: This change is needed because the test looks for a private field (the OFFSET field), and such field is declared in the typedef class' superclass. The old logic was generating things backwards, so we didn't have this issue.

test/jtreg/generator/nestedTypes/nested_types_names.h
```diff
@@ -0,0 +1,74 @@
...
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+// function declarations
+struct { int x; } f1(struct { int y; } p);
```
mcimadamore, line 25: This is a pretty convoluted test where we check declaration in parameter/return position. There were many issues in these cases, as the old jextract mostly didn't care about such structs (and that was not problematic, given that layouts were inlined all the time).

mcimadamore:
I've uploaded a new iteration which fixes name mangling for anon structs appearing in return/parameter position of a function/function pointer. In the old code we used to piggy back e.g. on the parameter name. E.g. if the parameter name is `p` and its type is an anon struct, the the anon struct has name `p`. I figured that this was too flimsy, as there can be many function parameters called `p`, and this will lead to a lot of mangled names e.g. `p$0`, `p$1`, etc.

The old jextract didn't have this issue as it did _not_ generate anything for such structs. But the new jextract needs these structs as it can refer to their layouts by name.

In the new naming scheme, given a function, or function pointer typedef with name `foo` we generate:
* `foo$x0`, `foo$x1` ... for anon structs associated with the parameter of the associated function type
* `foo$return` for the anon struct associated with the return of the associated function type

src/main/java/org/openjdk/jextract/impl/UnsupportedFilter.java
```diff
@@ -70,6 +70,8 @@ public Declaration.Scoped scan(Declaration.Scoped header) {
 
     @Override
     public Void visitFunction(Function funcTree, Declaration firstNamedParent) {
+        Utils.forEachNested(funcTree, s -> s.accept(this, firstNamedParent));
```
mcimadamore, line 73: Note: this method was not recursing into declarations nested in the return/parameters - this caused issues when a nested declaration happened to contain unsupported types, as `OutputFactory` would attempt to generate something for which the layout string cannot be generated.

test/jtreg/generator/outOfOrder/out_of_order_struct.h
```diff
@@ -0,0 +1,34 @@
...
+ * questions.
+ */
+
+struct Dummy {
+    struct tagFoo *Foo;
+};
+
+struct Bar {
+    int x;
+};
```
JornVernee, line 30: I think `Bar` is not strictly needed for this test case. (same for the typedef one).

JornVernee (commented):
This looks good. Great work on fixing such a tricky issue!

I just have one last concern. WRT:

> Note: libclang is not perfect here, as it often gives jextract duplicate cursors. That is, sometimes the same cursor can appear both at level N and at level N + 1 (this looks like a bug in libclang). The fact that jextract has builtin deduplication for declarations work out in our favor here.

Are we sure we're filtering out the right duplicates? e.g. if we filter out the N + 1 definition, doesn't the actual struct class we generate appear at the wrong level?

mcimadamore:
> 
> Are we sure we're filtering out the right duplicates? e.g. if we filter out the N + 1 definition, doesn't the actual struct class we generate appear at the wrong level?

My understanding here is that it doesn't matter much _which_ we retain, as long as (a) we have a consistent strategy to generate struct names based on what we keep (e.g. whether toplevel or "nested"), and (b) we don't end up generating things twice, redundantly. As a result, picking the "wrong" one can only result in a suboptimal name being generated (at least that's the theory).

src/main/java/org/openjdk/jextract/impl/NameMangler.java
```diff
@@ -196,24 +201,56 @@ public Void visitTypedef(Declaration.Typedef typedef, Declaration parent) {
...
+            default -> null;
+        };
+        if (func != null) {
+            // if this is a function pointer type def, try to use better fallback names for any
+            // anon struct/union that might be defined as part of this typedef
+            String suffix = null;
+            for (int i = 0 ; i < func.argumentTypes().size() ; i++) {
+                if (func.argumentTypes().get(i) instanceof Type.Declared declared && declared.tree() == nested) {
+                    // it's a function argument
+                    suffix = "$x" + i;
```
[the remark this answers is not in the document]
mcimadamore: In the general case no (e.g. variable whose type is a function pointer typedef). We might be able to improve in some selected case (but I'd prefer to address that separately).

JornVernee:
I tried to find a counter-example but it seems that in all cases where we would generate a nested class, e.g. for an inline struct decl in the type of a FieldDecl, clang doesn't have any duplicate top-level declarations. And in cases where there are duplicate top-level declarations (global vars and typedefs), we are generating top-level classes any way.

(Interestingly, it seems that if multiple inline declarations have the same name, Clang will treat subsequent declarations as anonymous. Clang seems to be quite strange here).

test/jtreg/generator/outOfOrder/out_of_order_typedef.h
```diff
@@ -0,0 +1,32 @@
...
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+typedef struct tagFoo Foo;
+
+struct Bar {
+    int x;
+};
```
JornVernee, line 28: `Bar` is not needed here either.

JornVernee (approved):

JornVernee (approved):

PR #168, opened 2023-12-19
Refresh libclang

mcimadamore:
Following the latest jextract changes, it is now time to refresh the libclang classes used by jextract. As a result, this change will drop `RuntimeHelper` and all constant classes (as the new generation strategy no longer needs them).

I encountered a couple of issues when running the updating script, which prompted me to simplify the script, and to add some more documentation.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/168/head:pull/168` \
`$ git checkout pull/168`

`$ git checkout pull/168` \
`$ git pull https://git.openjdk.org/jextract.git pull/168/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 168`

`$ git pr show -t 168`

</details>

<a href="https://git.openjdk.org/jextract/pull/168.diff">https://git.openjdk.org/jextract/pull/168.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/168#issuecomment-1862604201)


updateclang/extract.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+echo "Extracting libclang headers..."
```
mcimadamore, line 3: I've added these output strings so that it's possible to see what's going on

updateclang/extract.sh
```diff
@@ -8,7 +10,11 @@ jextract --output ../src/main/java \
   @clang.symbols \
   ${LIBCLANG_HOME}/include/clang-c/Index.h
 
+echo "Adding copyrights..."
+
 for x in ../src/main/java/org/openjdk/jextract/clang/libclang/*.java; do
-head -$COPYRIGHTLEN $x | diff cp_header.txt - || ( ( cat cp_header.txt; echo; cat $x) > /tmp/file;
-mv /tmp/file $x )
+(cat cp_header.txt; echo; cat $x) > /tmp/file;
```
mcimadamore, line 16: on my machine, the script hang here. In reality we don't need this complex logic, so I've tweaked the instructions so that the libclang folder is always deleted first (this would also ensure that there's no files leftover from previous runs).

updateclang/clang.symbols
```diff
@@ -13,8 +13,11 @@
 
 ###
 
+--include-typedef CXCursor
 --include-typedef CXCursorVisitor
 --include-typedef CXIndex
+--include-typedef CXSourceRange
```
mcimadamore, line 19: Interesting side-effect of the new translation strategy: since struct A depends on struct B in its layout, if you need to generate A now you also need to generate B. Otherwise the generated code will have an unresolved reference to B from A's layout.

JornVernee (approved):

PR #169, opened 2023-12-21
7903615: Jextract should generate javadoc contents using libclang pretty printer

mcimadamore:
This PR removes CDeclarationPrinter and, instead, uses libclang pretty-printing capabilities to print the javadoc contents for a given declaration.

The new logic will compute the comment string at parse time, in `TreeMaker`. The comment is then attached to the declaration using an attribute (and later retrieved when it's time to generate code).

I've used some custom printing policy options, after testing several of them to try and find the best results. With these options, there are two main differences from what we used to generate before:

1. named structs/unions are re-expanded in function declarations (e.g. the javadoc of a function accepting a `struct Foo` will also contain the definition of `Foo`)
2. there is no pretty-printing equivalent for types. This means that, for functional interfaces, we instead print the "parent" declaration, e.g. the declaration whose type is a function pointer.

It would be nice if we could improve on (1) by adding a new policy e.g. `IncludeAnonymousTagDefinition` - e.g. it is useful to have tag definitions inlined when the definition is anonymous (otherwise clang will generate a blob of text containing the source location of the anon definition), but it would be better perhaps not to inline the definition if it's not anonymous.

Example:

```
struct Point { int x; int y; };

void func(struct Point *pp);
```

This generates the following strings:

```
struct Point {
    int x;
    int y;
}

void func(struct Point {
    int x;
    int y;
} *pp)
```

I don't think it's a deal breaker, but it does make the javadoc more verbose.

Overall the changes are quite straightforward, and the code is simplified considerably. Also, this fixes a lot of little issues that arise when we try to generate javadoc for anonymous structs for which before only `struct` was printed.

There are only two situations in which I had to manually improve the clang output:

* macros - in this case pretty printer prints nothing :-) so I use the macro tokens to generate the corresponding doc string;
* enum constant - in this case pretty printer only prints the enum constant name, but the previous output had also the enum name and the value. I tweaked enum constants manually (also in `TreeMaker`). By doing that, I could then also remove `EnumConstantLifter`.

---------
### Progress

### Issue
 * [CODETOOLS-7903615](https://bugs.openjdk.org/browse/CODETOOLS-7903615): Jextract should generate javadoc contents using libclang pretty printer (**Enhancement** - P3)

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to [2d2fd9b7](https://git.openjdk.org/jextract/pull/169/files/2d2fd9b7b68fa95eaebc59a3c476da208c74f97b)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/169/head:pull/169` \
`$ git checkout pull/169`

`$ git checkout pull/169` \
`$ git pull https://git.openjdk.org/jextract.git pull/169/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 169`

`$ git pr show -t 169`

</details>

<a href="https://git.openjdk.org/jextract/pull/169.diff">https://git.openjdk.org/jextract/pull/169.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/169#issuecomment-1866260456)


mcimadamore:
I have added a very simple heuristic which should work relatively well in the majority of cases. Basically, now we first compute the cursor string w/o expanding struct/union/enum tags. Then, we look at the output: if the output contains strings such as `unnamed union at foo.h:9:11`, then we rollback, and compute another string where we expand all tags. Otherwise, we keep the string. In the vast majority of cases, this should do the right thing - the instances in which an anonymous tag is emitted are quite cornery (e.g. a function returning a struct defined on the fly). Note that anonymous structs are already expanded by the pretty printer, so if in C you have something like:

```
struct Foo {
    struct {
         int x;
     }
}
```

This will be rendered "as is" by the printer, no `<unnamed>` string will be emitted.

src/main/java/org/openjdk/jextract/impl/TreeMaker.java
```diff
@@ -449,4 +461,40 @@ private void checkCursorAny(Cursor c, CursorKind... kinds) {
         }
         throw new IllegalArgumentException("Invalid cursor kind");
     }
+
+    private <D extends Declaration> D withDeclarationString(D decl, Cursor cursor) {
+        String declString = switch (decl) {
+            case Declaration.Constant _ -> null; // do nothing for enum constants
+            case Typedef _ -> declarationString(cursor, true); // always expand typedefs
```
mcimadamore, line 468: typedefs are always printed in the expanded form because (a) it's not too annoying anyway and (b) in the more compact form, if the entity being typedeffed is an anonymous struct/union/enum, the output of the pretty printer can be weird. For instance in this case:

```
typedef struct { int x; int y; } Y;
```

The pretty printer generates:

```
typedef struct Y Y
```

Which is not great. The expanded form instead prints:

```
typedef struct {
    int x;
    int y;
} Y
```

as expected.

mcimadamore:
When testing with windows.h, I encountered a case where a macro constant had some comments in it, e.g.

```
#define FOO 42 /* a value */
```

Unfortunately, since my initial approach just copied all the tokens into the resulting javadoc, this fails. I've uploaded a fix for this where now macro values are generated in a much simpler fashion (w/o blindly copying the tokens). With this fix, windows.h is back to normal.

src/main/java/org/openjdk/jextract/impl/TreeMaker.java
```diff
@@ -226,7 +231,17 @@ public Declaration.Function createFunction(Cursor c) {
...
     public Declaration.Constant createMacro(Position pos, String name, Type type, Object value) {
-        return Declaration.constant(pos, name, value, type);
+        Declaration.Constant macro = Declaration.constant(pos, name, value, type);
+        String valueString = value.toString();
+        if (value instanceof String) {
+            // quote string literal
+            valueString = STR."\"\{valueString}\"";
+        } else if (Utils.isPointer(type)) {
+            // add pointer cast to make it look different from a numeric constant
+            valueString = STR."(\{valueString})";
```
JornVernee, line 241: Shouldn't this be something like:

```
valueString = STR."(\{type}) \{valueString}";
```

It seems like the current code would result in things like this:

    #define FOO (42)

src/main/java/org/openjdk/jextract/impl/TreeMaker.java
```diff
@@ -449,4 +468,40 @@ private void checkCursorAny(Cursor c, CursorKind... kinds) {
...
+                yield cursorString;
+            }
+        };
+        if (declString != null) {
+            DeclarationString.with(decl, declString);
+        }
+        return decl;
+    }
+
+    String declarationString(Cursor cursor, boolean expandNestedDecls) {
```
JornVernee, line 492: Seems like these 2 methods could be `private` as well.

src/main/java/org/openjdk/jextract/impl/TreeMaker.java
```diff
@@ -449,4 +468,40 @@ private void checkCursorAny(Cursor c, CursorKind... kinds) {
...
+            case Declaration.Constant _ -> null; // do nothing for enum constants
+            case Typedef _ -> declarationString(cursor, true); // always expand typedefs
+            default -> {
+                // heuristic, try w/o expanding first, and check if there are <anonymous> strings
+                String cursorString = declarationString(cursor, false);
+                if (cursorString.contains("\\(unnamed (struct|union|enum) at")) {
+                    // the output contains anonymous definitions, fallback and expand them
+                    cursorString = declarationString(cursor, true);
+                }
+                yield cursorString;
```
JornVernee, line 483: So, if I understand directly, this is different from what is described in the PR body. e.g. we will try to expand struct types as just  `struct Foo`, unless we also find anonymous decls, in which case we expand to `struct Foo { int x; int y; }` ?

mcimadamore: yes, this change was described in a follow up comment

src/main/java/org/openjdk/jextract/impl/TreeMaker.java
mcimadamore: I've fixed this by adding a cast to `void*` always. Note that `type.toString` doesn't return something that is similar to the C decl, so we can't really use that. (whether that's something we want to fix is a question for another PR I think)

PR #175, opened 2024-01-15
7903584: Revisit struct accessors

JornVernee:
Drop the `$get` and `$set` from the getter/setter names.

To ensure that we don't run into any clashes with the indexed accessors on structs (e.g. because a `long` field setter would clash with the indexed getter), we replace the indexed accessors with a 'asSlice' method which can be used to slice a given pointer using an index. This turns the following access pattern:

    MemorySegment arr = ...
    int x = Foo.x$get(arr, 10); // old 

Into this:

    MemorySegment arr = ...
    MemorySegment theFoo = Foo.asSlice(arr, 10);
    int x = Foo.x(theFoo); // new 

This then allows us to drop the indexed accessors altogether, thus avoiding a name clash.

---

This patch also drops the getters (both fro globals and struct fields) which return instances of a functional interface. There is an issue that the new getter name will clash with the FI instance getter (which was already using the nice name), but when discussing this with Maurizio we realized that returning wrappers, like these FI instances, is not really in the spirit of the current 'all-static' approach that jextract takes. So, this patch drops the FI instance getters. For now, the samples will manually wrap the address in an FI instance, until we address: https://bugs.openjdk.org/browse/CODETOOLS-7903628

Note: I did not test the LibffmpegMain and PanamaTime changes.

---------
### Progress

### Issues
 * [CODETOOLS-7903584](https://bugs.openjdk.org/browse/CODETOOLS-7903584): Revisit struct accessors (**Enhancement** - P3)
 * [CODETOOLS-7903417](https://bugs.openjdk.org/browse/CODETOOLS-7903417): Struct array slice accessor (**Enhancement** - P4)

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - **Reviewer**) ⚠️ Review applies to [d527b3f7](https://git.openjdk.org/jextract/pull/175/files/d527b3f7b04fe4577487e4ae1240bff3cdbef2fe)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/175/head:pull/175` \
`$ git checkout pull/175`

`$ git checkout pull/175` \
`$ git pull https://git.openjdk.org/jextract.git pull/175/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 175`

`$ git pr show -t 175`

</details>

<a href="https://git.openjdk.org/jextract/pull/175.diff">https://git.openjdk.org/jextract/pull/175.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/175#issuecomment-1892374674)

### DIFF
--- a/samples/libcurl/compile_windows.ps1
+++ b/samples/libcurl/compile_windows.ps1
@@ -9,7 +9,7 @@ jextract `
   --dump-includes 'includes_all.conf' `
   "$curlpath\include\curl\curl.h"
   
-Select-String -Path 'includes_all.conf' -Pattern 'curl' | %{ $_.Line } | Out-File -FilePath 'includes_filtered.conf' -Encoding ascii
+Select-String -Path 'includes_all.conf' -Pattern '(curl|sockaddr )' | %{ $_.Line } | Out-File -FilePath 'includes_filtered.conf' -Encoding ascii
 
 jextract `
   -t org.jextract `
--- a/src/main/java/org/openjdk/jextract/impl/StructBuilder.java
+++ b/src/main/java/org/openjdk/jextract/impl/StructBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
--- a/src/main/java/org/openjdk/jextract/impl/StructBuilder.java
+++ b/src/main/java/org/openjdk/jextract/impl/StructBuilder.java
@@ -88,6 +88,7 @@ final class StructBuilder extends ClassSourceBuilder implements OutputFactory.Bu
 
     void end() {
         if (!inAnonymousNested()) {
+            emitAt();
             appendBlankLine();
             emitSizeof();
             emitAllocatorAllocate();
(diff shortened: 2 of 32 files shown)


samples/libcurl/compile_windows.ps1
```diff
@@ -9,7 +9,7 @@ jextract `
   --dump-includes 'includes_all.conf' `
   "$curlpath\include\curl\curl.h"
   
-Select-String -Path 'includes_all.conf' -Pattern 'curl' | %{ $_.Line } | Out-File -FilePath 'includes_filtered.conf' -Encoding ascii
+Select-String -Path 'includes_all.conf' -Pattern '(curl|sockaddr )' | %{ $_.Line } | Out-File -FilePath 'includes_filtered.conf' -Encoding ascii
```
JornVernee, line 12: This was an existing issue that went unnoticed previously. The curl_sockaddr type depends on the Windows sockaddr type now (due to the layout referencing it directly), so we need to include that as well.

JornVernee:
Open question: should we add another overload of `asSlice` that takes a size of the returned segment? (right now the returned slice just goes up to the end of the given segment). If we do add an overload, should it accept a byte size or element count?

src/main/java/org/openjdk/jextract/impl/StructBuilder.java
```diff
@@ -200,6 +187,15 @@ private void emitSegmentGetter(String javaName, String offsetField, String sizeF
             """);
     }
 
+    private void emitAsSlice() {
+        appendIndentedLines(STR."""
+
+            public static MemorySegment asSlice(MemorySegment ptr, long index) {
```
mcimadamore, line 193: this might clash, right? Perhaps we can detect clashes locally, e.g. inside StructBuilder?

test/jtreg/generator/test8245003/Test8245003.java
```diff
@@ -78,7 +78,7 @@ public void testArrayAccessor() {
 
         seg = foo$SEGMENT();
         assertEquals(seg.byteSize(), Foo.sizeof());
-        assertEquals(Foo.count$get(seg), 37);
+        assertEquals(Foo.count(seg), 37);
         var greeting = Foo.greeting$slice(seg);
```
mcimadamore, line 82: should this be renamed too? E.g. if there's a struct field, just generate a segment getter with the same name of the filed?

JornVernee: I wasn't sure here, since the semantics are different. We originally decided to name this method `$slice` instead of `$get` after all.

I'm fine with going for the simpler name.

mcimadamore (approved):
Looks good (we can probably tweak the name clash if the needs arise in a separate PR)

JornVernee:
> (we can probably tweak the name clash if the needs arise in a separate PR)

Ok, I'll leave it like this then.

PR #176, opened 2024-01-15
7903627: Add an option to emit tracing calls to generated bindings

mcimadamore:
This PR tweaks the generated code to introduce a `TRACE_DOWNCALL` boolean flag in the generated code. This flag is initialized with the value of the runtime property `jextract.trace.downcalls`, and allows the generated code to print the value of arguments passed to downcalls. This can be useful when debugging.

Note: while generally the output provided is similar to that of `strace`, we have decided not to print the result value for now. That is because to do that, we'd need to first print the argument values (before the native call takes place) and then come back and print the result value (after the native call has returned). But this idiom would lead to messy output when multiple threads are used.

In `strace`, this is handled by buffering output, and only emitting something when the system call returns - but this is not an option here, as the tracing support is mostly intended to be used when diagnosing application crashes, in which case the native call never returns.

---------
### Progress

### Issue
 * [CODETOOLS-7903627](https://bugs.openjdk.org/browse/CODETOOLS-7903627): Add an option to emit tracing calls to generated bindings (**Enhancement** - P3)

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to [c1871e25](https://git.openjdk.org/jextract/pull/176/files/c1871e255b2aa57b6503c85d3df57e0937105a75)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/176/head:pull/176` \
`$ git checkout pull/176`

`$ git checkout pull/176` \
`$ git pull https://git.openjdk.org/jextract.git pull/176/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 176`

`$ git pr show -t 176`

</details>

<a href="https://git.openjdk.org/jextract/pull/176.diff">https://git.openjdk.org/jextract/pull/176.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/176#issuecomment-1892624587)


README.md
```diff
@@ -176,3 +176,37 @@ jextract -t org.jextract --source @includes.txt point.h
 
 It is easy to see how this mechanism allows developers to look into the set of symbols seen by `jextract` while parsing, and then process the generated include file, so as to prevent code generation for otherwise unused symbols.
 
+#### Tracing support
+
+It is sometimes useful to inspect the parameters passed to a native call, especially when diagnosing application
+bugs and/or crashes. The code generated by the `jextract` tool supports *tracing* of native calls, that is, parameters
+passed to native calls can be printed on the standard output.
+
+To enable the tracing support, just pass set the `-Djextract.trace.downcalls=true` flag to the launcher used to start the application.
```
JornVernee, line 185:
```suggestion
To enable the tracing support, just pass the `-Djextract.trace.downcalls=true` flag to the launcher used to start the application.
```

JornVernee (approved):
Maybe we should bump copyright years as well.

PR #178, opened 2024-01-16
7903626: Add struct helper method to reintrepret an array of structs

JornVernee:
This patch renames the existing `ofAddress` method we generate in struct classes to `reinterpret` to reflect the latest naming changes in the FFM API. It also adds another overload that takes an element count, for reinterpreting arrays. Finally, this patch adds the missing cleanup parameter that the underlying reinterpret method accepts.

---------
### Progress

### Issue
 * [CODETOOLS-7903626](https://bugs.openjdk.org/browse/CODETOOLS-7903626): Add struct helper method to reintrepret an array of structs (**Enhancement** - P3)

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - **Reviewer**) ⚠️ Review applies to [196c3b4b](https://git.openjdk.org/jextract/pull/178/files/196c3b4b2e1fa29867315b78b20a17d99f66c36d)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/178/head:pull/178` \
`$ git checkout pull/178`

`$ git checkout pull/178` \
`$ git pull https://git.openjdk.org/jextract.git pull/178/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 178`

`$ git pr show -t 178`

</details>

<a href="https://git.openjdk.org/jextract/pull/178.diff">https://git.openjdk.org/jextract/pull/178.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/178#issuecomment-1893943042)

### DIFF
--- a/src/main/java/org/openjdk/jextract/impl/StructBuilder.java
+++ b/src/main/java/org/openjdk/jextract/impl/StructBuilder.java
@@ -221,7 +221,7 @@ final class StructBuilder extends ClassSourceBuilder implements OutputFactory.Bu
         appendIndentedLines("""
 
             public static MemorySegment reinterpret(MemorySegment addr, Arena scope, Consumer<MemorySegment> cleanup) {
-                return addr.reinterpret($LAYOUT().byteSize(), scope, cleanup);
+                return reinterpret(addr, 1, scope, cleanup);
             }
 
             public static MemorySegment reinterpret(MemorySegment addr, long elementCount, Arena scope, Consumer<MemorySegment> cleanup) {
--- a/src/main/java/org/openjdk/jextract/impl/StructBuilder.java
+++ b/src/main/java/org/openjdk/jextract/impl/StructBuilder.java
@@ -92,7 +92,7 @@ final class StructBuilder extends ClassSourceBuilder implements OutputFactory.Bu
             emitSizeof();
             emitAllocatorAllocate();
             emitAllocatorAllocateArray();
-            emitOfAddressScoped();
+            emitReinterpret();
             classEnd();
             if (isNested()) {
                 // we are nested. Decrease align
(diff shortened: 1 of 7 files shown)


src/main/java/org/openjdk/jextract/impl/StructBuilder.java
```diff
@@ -217,11 +217,15 @@ public static MemorySegment allocateArray(long len, SegmentAllocator allocator)
...
     }
 
-    private void emitOfAddressScoped() {
+    private void emitReinterpret() {
         appendIndentedLines("""
 
-            public static MemorySegment ofAddress(MemorySegment addr, Arena scope) {
-                return addr.reinterpret($LAYOUT().byteSize(), scope, null);
+            public static MemorySegment reinterpret(MemorySegment addr, Arena scope, Consumer<MemorySegment> cleanup) {
+                return addr.reinterpret($LAYOUT().byteSize(), scope, cleanup);
```
mcimadamore, line 224: Maybe crazy idea - if we did `reinterpret(Long.MAX_VALUE, ...).asSlice($LAYOUT())` that would check alignment too - not sure if that would be useful or not.

mcimadamore: Should the simpler reinterpret method call the array one with length of 1 ?

test/TEST.groups
```diff
@@ -11,7 +11,7 @@ native_tests = \
     jtreg/generator/test8249757/LibTest8249757Test.java \
     jtreg/generator/test8252016/Test8252016.java \
     jtreg/generator/test8252121/Test8252121.java \
-    jtreg/generator/test8253102/LibTest8253102Test.java \
+    jtreg/generator/reinterpret/TestReinterpret.java \
```
mcimadamore, line 14: uhmmm - I did not realized we had to update this manually (for the other PRs...)

JornVernee: Intellij saved me here and caught it here too.

src/main/java/org/openjdk/jextract/impl/StructBuilder.java
JornVernee: > Should the simpler reinterpret method call the array one with length of 1 ?

yes, good suggestion

mcimadamore (approved):

src/main/java/org/openjdk/jextract/impl/StructBuilder.java
JornVernee: As for using asSlice as an alignment check. We don't have an asSlice method in MS which takes an element count, so we would also have to create a SequenceLayout on the fly. Not sure what to do... kinda feels like we need a reinterpret overload that takes an alignment, or another more convenient way to check alignment.

I think in principle checking alignment here is a good idea, since we're 'sanitizing' a raw address here, and an alignment check is a good way to catch corrupt pointers.

JornVernee: I'll leave it for now.

PR #181, opened 2024-01-16
7903632: struct accessor info is not derived from toplevel layout decl

mcimadamore:
This PR dials back the code generation updates a little when it comes to struct getter/setter, so that both their layouts and offsets is now derived from the toplevel struct layout - using layout paths, as before.

This allows clients to e.g. tweak layouts in a single place (the topelevel struct layout declaration) to adjust things such as endianness.

---------
### Progress

### Issue
 * [CODETOOLS-7903632](https://bugs.openjdk.org/browse/CODETOOLS-7903632): struct accessor info is not derived from toplevel layout decl (**Bug** - P3)

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to [326be9ca](https://git.openjdk.org/jextract/pull/181/files/326be9ca77bcbf710e19c40b84287a89483c6029)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/181/head:pull/181` \
`$ git checkout pull/181`

`$ git checkout pull/181` \
`$ git pull https://git.openjdk.org/jextract.git pull/181/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 181`

`$ git pr show -t 181`

</details>

<a href="https://git.openjdk.org/jextract/pull/181.diff">https://git.openjdk.org/jextract/pull/181.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/181#issuecomment-1894071956)

### DIFF
--- a/src/main/java/org/openjdk/jextract/impl/StructBuilder.java
+++ b/src/main/java/org/openjdk/jextract/impl/StructBuilder.java
@@ -139,11 +139,29 @@ final class StructBuilder extends ClassSourceBuilder implements OutputFactory.Bu
             String sizeField = emitSizeFieldDecl(varTree);
             emitSegmentGetter(javaName, offsetField, sizeField);
         } else if (Utils.isPointer(varTree.type()) || Utils.isPrimitive(varTree.type())) {
-            emitFieldGetter(javaName, varTree, offsetField);
-            emitFieldSetter(javaName, varTree, offsetField);
+            String layoutField = emitLayoutFieldDecl(varTree);
+            emitFieldGetter(javaName, varTree, layoutField, offsetField);
+            emitFieldSetter(javaName, varTree, layoutField, offsetField);
         } else {
             throw new IllegalArgumentException(STR."Type not supported: \{varTree.type()}");
         }
+    }
+
+    private List<String> prefixNamesList() {
+        return nestedAnonDeclarations.stream()
+                .map(d -> AnonymousStruct.anonName((Declaration.Scoped)d))
+                .toList().reversed();
+    }
+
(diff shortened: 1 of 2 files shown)


src/main/java/org/openjdk/jextract/impl/StructBuilder.java
```diff
@@ -139,40 +139,58 @@ public void addVar(Declaration.Variable varTree) {
...
-    private void emitFieldGetter(String javaName, Declaration.Variable varTree, String offsetField) {
+    private void emitFieldGetter(String javaName, Declaration.Variable varTree, String layoutField, String offsetField) {
         String seg = safeParameterName("seg");
         Class<?> type = Utils.carrierFor(varTree.type());
         appendBlankLine();
         emitFieldDocComment(varTree, "Getter for field:");
         appendIndentedLines(STR."""
             public static \{type.getSimpleName()} \{javaName}(MemorySegment \{seg}) {
-                return \{seg}.get(\{layoutString(varTree.type())}, \{offsetField});
+                return \{seg}.get(\{layoutField}, \{offsetField});
```
mcimadamore, line 180: overall, not too sure whether using separate layout/offset here is any better than using a var handle... opinions?

JornVernee (approved):
I think not creating the VarHandle is probably a bit better. We still end up generating the var handle for the field layout any way (since it's used in the MS getter/setter), but there is more chance of sharing the VarHandle with other use sites, since we don't bake in any offset computations.

PR #186, opened 2024-01-18
7903633: Make source mode the default, and drop compiled mode

JornVernee:
Make the source output mode the default, and drop the class output mode. (See JBS issue for motivation).

This is fairly straightforward. I've touched up the `Options` class a bit as well, as I noticed there were some unused elements. For the tests:
- Merged `JtregJextractSources` and `JtregJextract`. Updated generator tests to only call the merged runner
- `JextractToolRunner::run` is now also responsible for compiling generated source files. Renamed it to `runAndCompile`, and made the output directory an explicit argument, so that the util method can pick it up and compile the generated sources. Note that this method is now also responsible for checking that jextract runs successfully (which is required if we want to compile the output).
- Added a `JextractToolRunner::runNoOutput` method which just runs jextract without trying to look at the generated files (e.g. for negative tests that look for CLI errors).
- Added a `JextractToolRunner::run` method again, for the 2 tests that ran jextract with `--source`.

Furthermore, now that we only generate source files, `Writer` can be simplified a lot. I've removed our dependency on `JavaFileObject`, and replaced it with a new `JavaSourceFile` record. Unfortunately, I could not remove the module dependency on java.compiler, since the `NameMangler` also uses `SourceVersion`. Does replacing JavaFileObject seem like the right move? It reduces our dependency on java.compiler (maybe we could remove it completely at some point), and replaces `JavaFileObject` with a simpler API.

---------
### Progress

### Issue
 * [CODETOOLS-7903633](https://bugs.openjdk.org/browse/CODETOOLS-7903633): Make source mode the default, and drop compiled mode (**Bug** - P3)

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - **Reviewer**) ⚠️ Review applies to [098db9b3](https://git.openjdk.org/jextract/pull/186/files/098db9b39dc334dd1070837d695168cdff366340)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/186/head:pull/186` \
`$ git checkout pull/186`

`$ git checkout pull/186` \
`$ git pull https://git.openjdk.org/jextract.git pull/186/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 186`

`$ git pr show -t 186`

</details>

<a href="https://git.openjdk.org/jextract/pull/186.diff">https://git.openjdk.org/jextract/pull/186.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/186#issuecomment-1898571056)


src/main/java/org/openjdk/jextract/impl/Writer.java
```diff
@@ -26,64 +26,28 @@
 
 package org.openjdk.jextract.impl;
 
-import javax.tools.JavaFileObject;
+import org.openjdk.jextract.JavaSourceFile;
```
mcimadamore, line 29: Do we really need a class for this? Seems more like an helper method to the main jextract tool?

mcimadamore:
Regarding NameMangler - we also have these methods:
* https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Character.html#isJavaIdentifierStart(char)
* https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Character.html#isJavaIdentifierPart(char)

This is what the method in SourceVersion uses. Since NameMangler seems to copy other code from the javac sources, perhaps we can copy that predicate too, and drop the dependency - that method has not been updated in ages.

```
public static boolean isIdentifier(CharSequence name) {
        String id = name.toString();

        if (id.length() == 0) {
            return false;
        }
        int cp = id.codePointAt(0);
        if (!Character.isJavaIdentifierStart(cp)) {
            return false;
        }
        for (int i = Character.charCount(cp);
                i < id.length();
                i += Character.charCount(cp)) {
            cp = id.codePointAt(i);
            if (!Character.isJavaIdentifierPart(cp)) {
                return false;
            }
        }
        return true;
    }
    ```

File: test/lib/testlib/TestUtils.java [hunk 2]
```diff
@@ -48,6 +55,36 @@ public static Loader classLoader(Path... paths) {
...
+
+        StringWriter writer = new StringWriter();
+        PrintWriter pw = new PrintWriter(writer);
+
+        try {
+            System.err.println("compiling jextracted sources @ " + sourcePath.toAbsolutePath());
+            List<String> commands = new ArrayList<>();
+            commands.add("-parameters");
+            commands.add("--source=22");
+            commands.add("--enable-preview");
```
mcimadamore, line 74: I don't think the preview flag should be required? (since the API is now final)

src/main/java/org/openjdk/jextract/impl/CodeGenerator.java
```diff
@@ -25,16 +25,16 @@
...
 import java.io.PrintWriter;
 import java.util.List;
 import java.util.stream.Stream;
-import javax.tools.JavaFileObject;
 
 public final class CodeGenerator {
     private CodeGenerator() {}
 
-    public static JavaFileObject[] generate(Declaration.Scoped decl, String headerName,
+    public static JavaSourceFile[] generate(Declaration.Scoped decl, String headerName,
```
mcimadamore, line 37: Also this looks more like a method in JextractTool

mcimadamore (commented):
Looks good! Great improvements, and nice simplification of tests. Since the compilation we performed was just standard Java compilation, there's no need to add builtin compilation logic in jextract.

nlisker (changes requested):
The README should also be updated to remove the `--source` option. The places I found are:

* `jextract --source -t org.jextract point.h`
* Command line options table (there's also a problem with the macro description entry)
* `jextract -t org.jextract --source @includes.txt point.h`

test/jtreg/generator/clinitCycles/TestGlobal.java
```diff
@@ -29,19 +29,12 @@
 import java.lang.foreign.ValueLayout;
 
 /*
- * @test id=classes
+ * @test id
```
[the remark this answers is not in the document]
JornVernee: Right, `id` should be dropped too

src/main/java/org/openjdk/jextract/impl/Writer.java
JornVernee: You mean remove `Writer`? Or `JavaSourceFile`?

JornVernee:
@mcimadamore Note that besides `isIdentifier`, we also use `isKeyword` from SourceVersion. We could copy that as well, but it seems more likely to change.

JornVernee:
@nlisker Thanks for catching that. I'll have a look at the README as well

src/main/java/org/openjdk/jextract/impl/Writer.java
mcimadamore: `Writer`

sundararajana:
While you need java.compiler (because of SourceVersion.java dependency)

"--add-modules=org.openjdk.jextract,jdk.compiler,jdk.zipfs" may be changed Gradle file to remove jdk.compiler, jdk.zipfs.

mcimadamore:
> @mcimadamore Note that besides `isIdentifier`, we also use `isKeyword` from SourceVersion. We could copy that as well, but it seems more likely to change.

Uhm - true, I've missed that. Then it's probably better to keep the dependency on javax/lang/model

JornVernee:
I've pushed fixes for the review comments. Note that for the README, I've also pulled in a change from t he master branch which moves the build instructions to the end of the README. It looks like that change never made it into the `panama` branch.

JornVernee:
See the updates here: https://github.com/JornVernee/jextract/compare/2ccae1c..098db9b

README.md
```diff
@@ -200,3 +152,61 @@ glRotatef(0.0, 0.0, 1.0, 0.0)
 glutSolidTeapot(0.5)
 ```
 
+---
+
+### Building & Testing
```
mcimadamore, line 157: We have Building & Testing - and then Testing again?` I suggest maybe renaming this to `Building jextract` (for symmetry) and then `Testing jextract`.

src/main/java/org/openjdk/jextract/JextractTool.java
```diff
@@ -117,26 +121,64 @@ public static Declaration.Scoped parse(List<Path> headers, String... parserOptio
...
     }
 
     private static List<JavaSourceFile> generateInternal(Declaration.Scoped decl, String headerName,
                                                          String targetPkg, IncludeHelper includeHelper,
                                                          List<String> libNames, PrintWriter errStream) {
-        return List.of(CodeGenerator.generate(decl, headerName, targetPkg, includeHelper, libNames, errStream));
+        return List.of(generate(decl, headerName, targetPkg, includeHelper, libNames, errStream));
+    }
+
+    private static JavaSourceFile[] generate(Declaration.Scoped decl, String headerName,
```
mcimadamore, line 133: Thanks - I think seeing everything here helps a bit - as opposed to having to chase multiple classes around which only contained a single method.

mcimadamore (approved):

PR #190, opened 2024-01-22
Use regex match when looking for unnamed struct/union/enum in doc comment

JornVernee:
Noticed this small issue when running the samples. We try to look for an anonymous struct in the doc comments we generate. Clang will put the file path of the file containing the struct in the declaration string, which causes all kinds of problems, so we filter it out. A regular expression is used for this, however, it is passed to `String::contains`, which doesn't accept a regex. Switch to `String::matches` so it works.

- I've added a new test case to TestDocComments
- I've also renamed the enclosing folder of this test, since it still had the bug id as a name (while the test itself doesn't)
- I also noticed an issue in the TestGlobalStructAccess test. We access a global variable with a struct type, but this variable is not marked as exported, so the test fails on Windows.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/190/head:pull/190` \
`$ git checkout pull/190`

`$ git checkout pull/190` \
`$ git pull https://git.openjdk.org/jextract.git pull/190/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 190`

`$ git pr show -t 190`

</details>

<a href="https://git.openjdk.org/jextract/pull/190.diff">https://git.openjdk.org/jextract/pull/190.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/190#issuecomment-1903942039)


test/testng/org/openjdk/jextract/test/toolprovider/docComments/TestDocComments.java
```diff
@@ -169,6 +169,20 @@ public void testStruct2() throws IOException {
...
+        var comments = getDocComments("structs.h", "NestedAnon.java");
+        assertEquals(comments, List.of(
+            "struct NestedAnon { struct { int l; long long h; } u; }",
+            "struct { int l; long long h; }",
+            "Getter for field: int l",
+            "Setter for field: int l",
+            "Getter for field: long long h",
+            "Setter for field: long long h",
+            "Getter for field: struct { int l; long long h; } u",
+            "Setter for field: struct { int l; long long h; } u"));
```
JornVernee, line 183: Note that these last 2 comments would contain file paths without the fix.

When I was extracting Windows.h, some file paths contain `*\um\*` (which is an include directory in the Windows SDK), and this leads javac to complain about invalid unicode escape sequences.

mcimadamore (approved):
Thanks for fixing

PR #191, opened 2024-01-22
7903637: Use holder classes for string constants

JornVernee:
Use holder class idiom for string constants (See JBS issue for motivation).

The code in `emitConstant` is a bit copy-pasty. We need to strike a balance between giving a clear picture of the generated code by using a single string template as much as possible, and reducing duplication by splitting the templates up and weaving in more control flow.

I went with the repeated, but clearer approach here of having 2 separate templates. Let me know if you'd like me to change that.

---------
### Progress

### Issue
 * [CODETOOLS-7903637](https://bugs.openjdk.org/browse/CODETOOLS-7903637): Use holder classes for string constants (**Enhancement** - P4)

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - **Reviewer**) ⚠️ Review applies to [51d3e11a](https://git.openjdk.org/jextract/pull/191/files/51d3e11ac344a8842a5039c1a37b226dfead62db)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/191/head:pull/191` \
`$ git checkout pull/191`

`$ git checkout pull/191` \
`$ git pull https://git.openjdk.org/jextract.git pull/191/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 191`

`$ git pr show -t 191`

</details>

<a href="https://git.openjdk.org/jextract/pull/191.diff">https://git.openjdk.org/jextract/pull/191.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/191#issuecomment-1904587361)

### DIFF
--- a/src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
+++ b/src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
@@ -394,16 +394,28 @@ class HeaderFileBuilder extends ClassSourceBuilder {
 
     private void emitConstant(Class<?> javaType, String constantName, Object value, Declaration declaration) {
         incrAlign();
-        appendLines(STR."""
-            private static final \{javaType.getSimpleName()} \{constantName} = \{constantValue(javaType, value)};
-
-            """);
-        emitDocComment(declaration);
-        appendLines(STR."""
-            public static \{javaType.getSimpleName()} \{constantName}() {
-                return \{constantName};
-            }
-            """);
+        boolean useHolderClass = value instanceof String;
+        if (useHolderClass) {
+            emitDocComment(declaration);
+            appendLines(STR."""
+                public static \{javaType.getSimpleName()} \{constantName}() {
+                    class Holder {
+                        static final \{javaType.getSimpleName()} \{constantName} = \{constantValue(javaType, value)};
+                    }
+                    return Holder.\{constantName};
+                }
+                """);
+        } else {
+            appendLines(STR."""
+                private static final \{javaType.getSimpleName()} \{constantName} = \{constantValue(javaType, value)};
+                """);
+            emitDocComment(declaration);
+            appendLines(STR."""
+                public static \{javaType.getSimpleName()} \{constantName}() {
+                    return \{constantName};
+                }
+                """);
+        }
         decrAlign();
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -393,22 +394,34 @@ class Holder {
...
+                }
+                """);
+        }
         decrAlign();
     }
 
     private String constantValue(Class<?> type, Object value) {
         if (value instanceof String) {
-            return STR."Arena.ofAuto().allocateFrom(\"\{Utils.quote(Objects.toString(value))}\")";
+            return STR."\{runtimeHelperName()}.LIBRARY_ARENA.allocateFrom(\"\{Utils.quote(Objects.toString(value))}\")";
```
mcimadamore, line 424: I think I'd prefer to inline this in the above method `emitConstant` - as we're not really supposed to call allocation methods from plain field initializers (e.g. this is only for holder class).

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -393,22 +394,34 @@ class Holder {
...
-            private static final \{javaType.getSimpleName()} \{constantName} = \{constantValue(javaType, value)};
-
-            """);
-        emitDocComment(declaration);
-        appendLines(STR."""
-            public static \{javaType.getSimpleName()} \{constantName}() {
-                return \{constantName};
-            }
-            """);
+        boolean useHolderClass = value instanceof String;
```
mcimadamore, line 397: maybe `isString` would be a better name

JornVernee: I'll just drop the intermediate variable

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -420,9 +420,7 @@ class Holder {
     }
 
     private String constantValue(Class<?> type, Object value) {
-        if (value instanceof String) {
-            return STR."\{runtimeHelperName()}.LIBRARY_ARENA.allocateFrom(\"\{Utils.quote(Objects.toString(value))}\")";
-        } else if (type == MemorySegment.class) {
+        if (type == MemorySegment.class) {
             return STR."MemorySegment.ofAddress(\{((Number)value).longValue()}L)";
         } else {
```
mcimadamore, line 425: Can we beef up this method with something like `if (value instanceof Number) <current logic> else throw UOE` ?

mcimadamore (approved):

PR #192, opened 2024-01-23
7903629: Emit allocateFrom method for struct classes

JornVernee:
Adds an `allocateFrom` method to struct classes, which can be used like a constructor to allocate and initialize a struct in one call.

I realized that, while this works well for initializing single structs, it doesn't work for arrays of structs. So, I've also added an `init` method which can be used to initialize a given struct. This allows a client to first allocate an array, and then initialize each element in a single shot (see test case). (we can then also just call `init` from `allocateFrom`s implementation).

Open question: I'm currently not emitting the new methods if the struct doesn't have any fields, since they are really not needed in that case. But, maybe `allocateFrom` and `init` should also be emitted for empty struct for the sake of consistency?

---------
### Progress

### Issue
 * [CODETOOLS-7903629](https://bugs.openjdk.org/browse/CODETOOLS-7903629): Emit allocateFrom method for struct classes (**Enhancement** - P3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/192/head:pull/192` \
`$ git checkout pull/192`

`$ git checkout pull/192` \
`$ git pull https://git.openjdk.org/jextract.git pull/192/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 192`

`$ git pr show -t 192`

</details>

<a href="https://git.openjdk.org/jextract/pull/192.diff">https://git.openjdk.org/jextract/pull/192.diff</a>

</details>

### DIFF
--- a/src/main/java/org/openjdk/jextract/impl/StructBuilder.java
+++ b/src/main/java/org/openjdk/jextract/impl/StructBuilder.java
@@ -51,6 +51,8 @@ final class StructBuilder extends ClassSourceBuilder implements OutputFactory.Bu
     private final Declaration.Scoped structTree;
     private final Type structType;
     private final Deque<Declaration> nestedAnonDeclarations;
+    private record Field(String name, Class<?> type) {}
+    private final List<Field> seenFields = new ArrayList<>();
 
     StructBuilder(SourceFileBuilder builder, String modifiers, String className,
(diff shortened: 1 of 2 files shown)


JornVernee:
Closing as 'won't fix' for now (See comment on JBS issue)

PR #195, opened 2024-01-25
7903644: typedef of anonymous struct generates a redundant class

mcimadamore:
This PR tweaks jextract so that typedefs of anonymous structs only generate one class (that of the struct).

This required some changes in both `NameMangler` (to make sure that the special naming rule was applied), and then in `OutputFactory`, to make sure that in such case the typedef class is not emitted.

I've tweaked a number of tests which now reflect the "less verbose" jextract output - as such I don't think we need to add further tests.

---------
### Progress

### Issue
 * [CODETOOLS-7903644](https://bugs.openjdk.org/browse/CODETOOLS-7903644): typedef of anonymous struct generates a redundant class (**Bug** - P3)

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to [726967aa](https://git.openjdk.org/jextract/pull/195/files/726967aaeb20caf025d1924e5d934570b80721fa)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/195/head:pull/195` \
`$ git checkout pull/195`

`$ git checkout pull/195` \
`$ git pull https://git.openjdk.org/jextract.git pull/195/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 195`

`$ git pr show -t 195`

</details>

<a href="https://git.openjdk.org/jextract/pull/195.diff">https://git.openjdk.org/jextract/pull/195.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/195#issuecomment-1910674183)

### DIFF
--- a/src/main/java/org/openjdk/jextract/impl/NameMangler.java
+++ b/src/main/java/org/openjdk/jextract/impl/NameMangler.java
@@ -172,7 +172,7 @@ public final class NameMangler implements Declaration.Visitor<Void, Declaration>
                 if (parent instanceof Typedef typedef && typedef.type() instanceof Declared declared &&
                         declared.tree().name().isEmpty()) {
                     // typedef struct { ... } Foo;
-                    // steal the name from the parent typedef
+                    // steal the name from the parent typedef (which has already been mangled)
                     name = JavaName.getOrThrow(parent);
                 } else {
                     name = oldScope.uniqueNestedClassName(scoped.name().isEmpty() ?
--- a/src/main/java/org/openjdk/jextract/impl/NameMangler.java
+++ b/src/main/java/org/openjdk/jextract/impl/NameMangler.java
@@ -28,6 +28,7 @@ import org.openjdk.jextract.Declaration;
 import org.openjdk.jextract.Declaration.Typedef;
 import org.openjdk.jextract.Declaration.Variable;
 import org.openjdk.jextract.Type;
+import org.openjdk.jextract.Type.Declared;
 import org.openjdk.jextract.Type.Delegated;
 import org.openjdk.jextract.Type.Function;
 import org.openjdk.jextract.impl.DeclarationImpl.AnonymousStruct;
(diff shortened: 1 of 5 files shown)


src/main/java/org/openjdk/jextract/impl/NameMangler.java
```diff
@@ -192,9 +200,16 @@ public Void visitTypedef(Declaration.Typedef typedef, Declaration parent) {
...
-        String javaName = curScope.uniqueNestedClassName(typedef.name());
+        String javaName;
+        if (Utils.getAnonymousStructTypedef(typedef) != null) {
+            // a typedef of a struct/union without a name is special-cased,
+            // so that the structs "inherits" the typedef name.
+            javaName = typedef.name();
+        } else {
+            // We may potentially generate a class for a typedef. Make sure
+            // class name is unique in the current nesting context.
+            javaName = curScope.uniqueNestedClassName(typedef.name());
```
mcimadamore, line 211: This logic is a bit subtle: we have to avoid calling `uniqueNestedClassName` in case there's a struct that will inherit the typedef name - otherwise the mangler will see _two_ attempts to add the typedef name to the current scope, which will result in extra mangling.

JornVernee: This seems a bit ad-hoc... Suggestion: the `Scope` constructor tries to generate a unique nested class name if `parent` is not `null`. That's only the case when calling `newStruct`, which is only called from `visitScoped`. So, I think we can move the logic that computes the nested class name from the `Scope` constructor to `visitScoped` (to the `else` branch this patch adds). Then we can just pass the unmangled name from the `if` branch to `newStruct`, and there should be no need to have this logic here.

JornVernee (approved):

src/main/java/org/openjdk/jextract/impl/NameMangler.java
mcimadamore: Thanks for the suggestion - I did think of something similar, but had not realized we only needed the mangling in one place. Updated the code, looks a bit better I think.

PR #198, opened 2024-01-29
7903649: Field and global variables of array type should have indexed accessors

mcimadamore:
This PR adds support for indexed accessors for struct fields and global variables whose type is an array type. These accessors feature a number of `long` access coordinates which has the same cardinality as that of the underlying array type.

For instance, consider the following global variable declaration:

```c
int ints[2][3][4];
```

For this, jextract now emits:

```java
    private static SequenceLayout ints$LAYOUT() {
        class Holder {
            static final SequenceLayout LAYOUT = MemoryLayout.sequenceLayout(2, MemoryLayout.sequenceLayout(3, MemoryLayout.sequenceLayout(4, foo_h.C_INT)));
        }
        return Holder.LAYOUT;
    }

    private static MemorySegment ints$SEGMENT() {
        class Holder {
            static final MemorySegment SEGMENT = foo_h.findOrThrow("ints")
                .reinterpret(ints$LAYOUT().byteSize());
        }
        return Holder.SEGMENT;
    }

    /**
     * Getter for variable:
     * {@snippet lang=c :
     * int ints[2][3][4]
     * }
     */
    public static MemorySegment ints() {
        return ints$SEGMENT();
    }

    /**
     * Setter for variable:
     * {@snippet lang=c :
     * int ints[2][3][4]
     * }
     */
    public static void ints(MemorySegment varValue) {
        MemorySegment.copy(varValue, 0L, ints$SEGMENT(), 0L, ints$LAYOUT().byteSize());
    }

    private static VarHandle ints$ELEM_HANDLE() {
        class Holder {
            static final VarHandle HANDLE = ints$LAYOUT().varHandle(sequenceElement(), sequenceElement(), sequenceElement());
        }
        return Holder.HANDLE;
    }

    /**
     * Indexed getter for variable:
     * {@snippet lang=c :
     * int ints[2][3][4]
     * }
     */
    public static int ints(long index0, long index1, long index2) {
        return (int)ints$ELEM_HANDLE().get(ints$SEGMENT(), 0L, index0, index1, index2);
    }

    /**
     * Indexed setter for variable:
     * {@snippet lang=c :
     * int ints[2][3][4]
     * }
     */
    public static void ints(long index0, long index1, long index2, int varValue) {
        ints$ELEM_HANDLE().set(ints$SEGMENT(), 0L, index0, index1, index2, varValue);
    }
```

If the array element type is a struct, different code needs to be generated. Consider this global variable declaration:

```c
struct Point { int x; int y; } points[2][3][4];
```

This generates the following:

```java
    private static SequenceLayout points$LAYOUT() {
        class Holder {
            static final SequenceLayout LAYOUT = MemoryLayout.sequenceLayout(2, MemoryLayout.sequenceLayout(3, MemoryLayout.sequenceLayout(4, Point.layout())));
        }
        return Holder.LAYOUT;
    }

    private static MemorySegment points$SEGMENT() {
        class Holder {
            static final MemorySegment SEGMENT = foo_h.findOrThrow("points")
                .reinterpret(points$LAYOUT().byteSize());
        }
        return Holder.SEGMENT;
    }

    /**
     * Getter for variable:
     * {@snippet lang=c :
     * struct Point points[2][3][4]
     * }
     */
    public static MemorySegment points() {
        return points$SEGMENT();
    }

    /**
     * Setter for variable:
     * {@snippet lang=c :
     * struct Point points[2][3][4]
     * }
     */
    public static void points(MemorySegment varValue) {
        MemorySegment.copy(varValue, 0L, points$SEGMENT(), 0L, points$LAYOUT().byteSize());
    }

    private static MethodHandle points$ELEM_HANDLE() {
        class Holder {
            static final MethodHandle HANDLE = points$LAYOUT().sliceHandle(sequenceElement(), sequenceElement(), sequenceElement());
        }
        return Holder.HANDLE;
    }

    /**
     * Indexed getter for variable:
     * {@snippet lang=c :
     * struct Point points[2][3][4]
     * }
     */
    public static MemorySegment points(long index0, long index1, long index2) {
        try {
            return (MemorySegment)points$ELEM_HANDLE().invokeExact(points$SEGMENT(), 0L, index0, index1, index2);
        } catch (Throwable ex$) {
        }
    }

    /**
     * Indexed setter for variable:
     * {@snippet lang=c :
     * struct Point points[2][3][4]
     * }
     */
    public static void points(long index0, long index1, long index2, MemorySegment varValue) {
        MemorySegment.copy(varValue, 0L, points(index0, index1, index2), 0L, Point.layout().byteSize());
    }
```

(note that bulk copy is used for the setter).

For struct fields, we generate similar code - the only difference is that the var/method handle for accessing the array element is declared in a plain field (instead of using the holder idiom), similarly to what happens with other struct declarations. Note also that, for structs, the indexed accessors must receive an additional `MemorySegment` parameter (the segment struct being accessed).

### DIFF
--- a/src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
+++ b/src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
@@ -39,6 +39,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.StringJoiner;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 /**
  * A helper class to generate header interface class in source form.
@@ -59,6 +61,13 @@ class HeaderFileBuilder extends ClassSourceBuilder {
             String segmentConstant = emitGlobalSegment(layoutVar, javaName, nativeName);
             emitGlobalSegmentGetter(segmentConstant, layoutVar, javaName, varTree, "Getter for variable:");
             emitGlobalSegmentSetter(segmentConstant, layoutVar, javaName, varTree, "Setter for variable:");
+            int dims = Utils.dimensions(varTree.type());
+            if (dims > 0) {
+                String arrayHandle = emitArrayElementHandle(javaName, varTree, layoutVar, dims);
+                IndexList indexList = IndexList.of(dims);
+                emitGlobalArrayGetter(segmentConstant, arrayHandle, indexList, javaName, varTree, "Indexed getter for variable:");
+                emitGlobalArraySetter(segmentConstant, arrayHandle, indexList, javaName, varTree, "Indexed setter for variable:");
+            }
         } else if (Utils.isPointer(varTree.type()) || Utils.isPrimitive(varTree.type())) {
             String segmentConstant = emitGlobalSegment(layoutVar, javaName, nativeName);
             emitGlobalGetter(segmentConstant, layoutVar, javaName, varTree, "Getter for variable:");
@@ -330,8 +339,8 @@ class HeaderFileBuilder extends ClassSourceBuilder {
         emitDocComment(decl, docHeader);
         Class<?> type = Utils.carrierFor(decl.type());
         appendLines(STR."""
-            public static void \{javaName}(\{type.getSimpleName()} x) {
-                \{segmentConstant}.set(\{layoutVar}, 0L, x);
+            public static void \{javaName}(\{type.getSimpleName()} varValue) {
+                \{segmentConstant}.set(\{layoutVar}, 0L, varValue);
             }
             """);
         decrAlign();
@@ -376,6 +385,86 @@ class HeaderFileBuilder extends ClassSourceBuilder {
             }
             """);
         return STR."\{mangledName}()";
+    }
+
+    private String emitArrayElementHandle(String javaName, Declaration.Variable varTree, String fieldLayoutName, int dims) {
+        String arrayHandleName = STR."\{javaName}$ELEM_HANDLE";
+        String path = IntStream.range(0, dims)
+                .mapToObj(_ -> "sequenceElement()")
+                .collect(Collectors.joining(", "));
+        Type elemType = Utils.typeOrElemType(varTree.type());
+        if (Utils.isStructOrUnion(elemType)) {
+            appendIndentedLines(STR."""
+
+                private static MethodHandle \{arrayHandleName}() {
+                    class Holder {
+                        static final MethodHandle HANDLE = \{fieldLayoutName}.sliceHandle(\{path});
+                    }
+                    return Holder.HANDLE;
+                }
+                """);
+        } else {
+            appendIndentedLines(STR."""
+
+                private static VarHandle \{arrayHandleName}() {
+                    class Holder {
+                        static final VarHandle HANDLE = \{fieldLayoutName}.varHandle(\{path});
+                    }
+                    return Holder.HANDLE;
+                }
+                """);
+        }
+        return STR."\{arrayHandleName}()";
+    }
+
+    private void emitGlobalArrayGetter(String segmentConstant, String arrayElementHandle, IndexList indexList,
+                                       String javaName, Declaration.Variable varTree, String docHeader) {
+        Type elemType = Utils.typeOrElemType(varTree.type());
+        Class<?> typeCls = Utils.carrierFor(elemType);
+        appendBlankLine();
+        incrAlign();
+        emitDocComment(varTree, docHeader);
+        if (Utils.isStructOrUnion(elemType)) {
+            appendLines(STR."""
+                public static MemorySegment \{javaName}(\{indexList.decl()}) {
+                    try {
+                        return (MemorySegment)\{arrayElementHandle}.invokeExact(\{segmentConstant}, 0L, \{indexList.use()});
+                    } catch (Throwable ex$) {
+                        throw new AssertionError("should not reach here", ex$);
+                    }
+                }
+                """);
+        } else {
+            appendLines(STR."""
+                public static \{typeCls.getSimpleName()} \{javaName}(\{indexList.decl()}) {
+                    return (\{typeCls.getSimpleName()})\{arrayElementHandle}.get(\{segmentConstant}, 0L, \{indexList.use()});
+                }
+                """);
+        }
+        decrAlign();
+    }
+
+    private void emitGlobalArraySetter(String segmentConstant, String arrayElementHandle, IndexList indexList,
+                                       String javaName, Declaration.Variable varTree, String docHeader) {
+        Type elemType = Utils.typeOrElemType(varTree.type());
+        Class<?> typeCls = Utils.carrierFor(elemType);
+        appendBlankLine();
+        incrAlign();
+        emitDocComment(varTree, docHeader);
+        if (Utils.isStructOrUnion(elemType)) {
+            appendLines(STR."""
+                public static void \{javaName}(\{indexList.decl()}, MemorySegment varValue) {
+                    MemorySegment.copy(varValue, 0L, \{javaName}(\{indexList.use()}), 0L, \{layoutString(elemType)}.byteSize());
+                }
+                """);
+        } else {
+            appendLines(STR."""
+                public static void \{javaName}(\{indexList.decl()}, \{typeCls.getSimpleName()} varValue) {
+                    \{arrayElementHandle}.set(\{segmentConstant}, 0L, \{indexList.use()}, varValue);
+                }
+                """);
+        }
+        decrAlign();
     }
 
     private String emitVarLayout(Type varType, String javaName) {
--- a/src/main/java/org/openjdk/jextract/impl/StructBuilder.java
+++ b/src/main/java/org/openjdk/jextract/impl/StructBuilder.java
@@ -42,6 +42,7 @@ import java.util.ArrayList;
 import java.util.Deque;
 import java.util.List;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 /**
  * This class generates static utilities class for C structs, unions.
@@ -133,13 +134,19 @@ final class StructBuilder extends ClassSourceBuilder implements OutputFactory.Bu
     public void addVar(Declaration.Variable varTree) {
         String javaName = JavaName.getOrThrow(varTree);
         appendBlankLine();
+        String layoutField = emitLayoutFieldDecl(varTree);
         String offsetField = emitOffsetFieldDecl(varTree);
         if (Utils.isArray(varTree.type()) || Utils.isStructOrUnion(varTree.type())) {
-            String sizeField = emitSizeFieldDecl(varTree);
-            emitSegmentGetter(javaName, varTree, offsetField, sizeField);
-            emitSegmentSetter(javaName, varTree, offsetField, sizeField);
+            emitSegmentGetter(javaName, varTree, offsetField, layoutField);
+            emitSegmentSetter(javaName, varTree, offsetField, layoutField);
+            int dims = Utils.dimensions(varTree.type());
+            if (dims > 0) {
+                String arrayHandle = emitArrayElementHandle(javaName, varTree, layoutField, dims);
+                IndexList indexList = IndexList.of(dims);
+                emitFieldArrayGetter(javaName, varTree, arrayHandle, indexList);
+                emitFieldArraySetter(javaName, varTree, arrayHandle, indexList);
+            }
         } else if (Utils.isPointer(varTree.type()) || Utils.isPrimitive(varTree.type())) {
-            String layoutField = emitLayoutFieldDecl(varTree);
             emitFieldGetter(javaName, varTree, layoutField, offsetField);
             emitFieldSetter(javaName, varTree, layoutField, offsetField);
         } else {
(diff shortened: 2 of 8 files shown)


test/jtreg/generator/arrayAccess/TestArrayAccess.java
```diff
@@ -0,0 +1,211 @@
...
+ * @build TestArrayAccess
+ * @run testng/othervm --enable-native-access=ALL-UNNAMED TestArrayAccess
+ */
+
+public class TestArrayAccess {
+    @Test
+    public void testArrayAccessStructInt1() {
+        try (Arena arena = Arena.ofConfined()) {
+            MemorySegment foo = Foo.allocate(arena);
+            for (int i = 0 ; i < 2 ; i++) {
```
JornVernee, line 48: So, the array length is hard-coded here. What would it take to derive the length from the bindings?

I think in this case it would be:

```
long count = Foo.layout().select(groupElement("ints1")).elementCount();
```

?

JornVernee (approved):

test/jtreg/generator/arrayAccess/TestArrayAccess.java
JornVernee: P.S. I guess what I'm implying here is: maybe we need a way to make getting the length of an array field easier as well

mcimadamore: Good observation... unfortunately I'm not too sure on how to address this - in the sense that a single field can have N length accessors (one per dimension) - well, or at least N static fields. We could simply expose the field layout (and leave it to the client to decode the dimensions) - but that's not too usable, and it also feels a bit redundant (since the field layout can be derived, as you show, from the struct layout). Also, for globals we ditched layouts (because of accessor naming troubles), so the same approach wouldn't work there.

It feels like doubling down on layouts is the right way to go - and it would be generally useful to maybe have a layout per field/global variable. But the logistics of it are complicated, especially for global variables (where you'd need a extra holder class for each global variable to store the layout), and naming can be a pain too.

mcimadamore: Thinking a bit more... while the hardwired array lengths are not ideal, I guess I still see the glass as half-full: the javadoc of the accessor shows the array declaration in full, and that will have the correct sizes displayed. Yes, it's a bit of a pity that the sizes are only captured in the javadoc, but I'm not sure we can fix that w/o moving to a more complex translation scheme. Also, the indices passed to the array accessor are validated dynamically (as the array handle is derived from the layout), so there's at least some dynamic safety.

If we want to capture the dimensions in the generated code, we need at least a single `long[]` array. But then the problem is with naming: if we call this `<fieldName>$dims`, we go back to dollar names (which we tried to back away from with recent changes). The biggest issue of using simpler names (as we do now) is that we don't really have a great story for adding extra accessors in a way that is easily discoverable for clients (we have this same issue for exposing layouts of global variables, or function descriptors of native functions).

The only half-smart idea I have in this respect is to add, for each symbol S, a static class with the same name which contains all the relevant constants. E.g.

```
class Foo_h {
     int x() { ... }
     void x(int val) { ... }
     static class x {
          static final LAYOUT = ...
          ...
      }
}
```

That is, make the holder class we generate _public_, so that it can be a placeholder for constants that can be useful to access from clients of the generated code. At least for the header class, this approach doesn't add any overhead, given that we emit an holder class per function/global variable _anyway_. For structs, we currently do not generate holder classes, so this would increase overhead in the generated code. But, perhaps, a more regular (even if dumber) translation scheme for _all_ variable/fields might result in more predictability/discoverability? (separately, it feels like the code generation for structs is doing quite a bit of stuff in its static initializers, like calling `MemoryLayout::byteOffset`, or `MemoryLayout::varHandle`, so perhaps moving this logic away into an holder class might be beneficial anyway).

mcimadamore: Here's a concrete example (obtained from quickly hacking jextract's generated output). Consider this C global var declaration:

```c
struct Point { int x; int y; } aPoint[3][4];
```

And here's the code generated by jextract:

 ```java
 public static class aPoint {
        static final SequenceLayout LAYOUT = MemoryLayout.sequenceLayout(3, MemoryLayout.sequenceLayout(4, Point.layout()));
        static final MemorySegment SEGMENT = foo_h.findOrThrow("aPoint").reinterpret(LAYOUT.byteSize());
        static final MethodHandle HANDLE = LAYOUT.sliceHandle(sequenceElement(), sequenceElement());
        static final long[] DIMS = { 3, 4 };
    }

    /**
     * Getter for variable:
     * {@snippet lang=c :
     * struct Point aPoint[3][4]
     * }
     */
    public static MemorySegment aPoint() {
        return aPoint.SEGMENT;
    }

    /**
     * Setter for variable:
     * {@snippet lang=c :
     * struct Point aPoint[3][4]
     * }
     */
    public static void aPoint(MemorySegment varValue) {
        MemorySegment.copy(varValue, 0L, aPoint.SEGMENT, 0L, aPoint.LAYOUT.byteSize());
    }

    /**
     * Indexed getter for variable:
     * {@snippet lang=c :
     * struct Point aPoint[3][4]
     * }
     */
    public static MemorySegment aPoint(long index0, long index1) {
        try {
            return (MemorySegment)aPoint.HANDLE.invokeExact(aPoint.SEGMENT, 0L, index0, index1);
        } catch (Throwable ex$) {
        }
    }

    /**
     * Indexed setter for variable:
     * {@snippet lang=c :
     * struct Point aPoint[3][4]
     * }
     */
    public static void aPoint(long index0, long index1, MemorySegment varValue) {
        MemorySegment.copy(varValue, 0L, aPoint(index0, index1), 0L, Point.layout().byteSize());
    }
```

The accessors are still as available and as discoverable as before. But if the clients want to know additional info, they can dig deeper, by accessing the static fields on the public holder class. Thoughts?

Comment (JornVernee) [reply test/jtreg/generator/arrayAccess/TestArrayAccess.java:48]: I think this looks great, at least for globals. For struct fields it might be okay as well, but it seems tempting to try and find a way to put these extra fields directly in the struct class.

Comment (mcimadamore) [reply test/jtreg/generator/arrayAccess/TestArrayAccess.java:48]: FTR, if the constant class is called `<varName>$constants` all the ambiguities issues are resolved, and all tests pass. But, we're back with dollar names :-)

If we're fine with dollar names for "less common API points", we could settle by having a single extra layout accessor/descriptor accessor per var/field/function. I think the layout API should expose an handy method for computing dimensions of a possibly nested sequence layout.

Comment (mcimadamore) [reply test/jtreg/generator/arrayAccess/TestArrayAccess.java:48]: > we could settle by having a single extra layout accessor/descriptor accessor per var/field/function

Although, I think I like the idea of exposing multiple related constants under a single umbrella - given that in most cases we generate the holder class anyway. In many cases there seem to be multiple interesting constants for a given symbol - and if we were to expose different constant accessors, the constant accessor would quickly outnumber the _real_ API point, which would then lead us back in a situation of poor API discoverability.

In other words, while reserving the "good name" to the common API points remains a good move, if such good names are buried under several "less than good names", the API quickly becomes bloated, which is still problematic, unfortunately.

Comment (mcimadamore) [reply test/jtreg/generator/arrayAccess/TestArrayAccess.java:48]: I've put together another branch where I fully explored the idea of exposing constants (for global vars, functions and struct fields) in a separate holder class:

https://github.com/openjdk/jextract/compare/panama...mcimadamore:jextract:public_holders?expand=1

For globals and functions, not much changes (as we needed an holder class even before this).

For structs, additional holder classes are generated (one per field). For instance, here's what `Point` looks like:

```java
/**
 * {@snippet lang=c :
 * struct Point {
 *     int x;
 *     int y;
 * }
 * }
 */
public class Point {

    Point() {
        // Should not be called directly
    }

    private static final GroupLayout $LAYOUT = MemoryLayout.structLayout(
        foo_h.C_INT.withName("x"),
        foo_h.C_INT.withName("y")
    ).withName("Point");

    /**
     * The layout of this struct
     */
    public static final GroupLayout layout() {
        return $LAYOUT;
    }

    public static class x$constants {
        public static final OfInt LAYOUT = (OfInt)layout().select(groupElement("x"));
        public static final long OFFSET = layout().byteOffset(groupElement("x"));
    }

    /**
     * Getter for field:
     * {@snippet lang=c :
     * int x
     * }
     */
    public static int x(MemorySegment struct) {
        return struct.get(x$constants.LAYOUT, x$constants.OFFSET);
    }

    /**
     * Setter for field:
     * {@snippet lang=c :
     * int x
     * }
     */
    public static void x(MemorySegment struct, int fieldValue) {
        struct.set(x$constants.LAYOUT, x$constants.OFFSET, fieldValue);
    }

    public static class y$constants {
        public static final OfInt LAYOUT = (OfInt)layout().select(groupElement("y"));
        public static final long OFFSET = layout().byteOffset(groupElement("y"));
    }

    /**
     * Getter for field:
     * {@snippet lang=c :
     * int y
     * }
     */
    public static int y(MemorySegment struct) {
        return struct.get(y$constants.LAYOUT, y$constants.OFFSET);
    }

    /**
     * Setter for field:
     * {@snippet lang=c :
     * int y
     * }
     */
    public static void y(MemorySegment struct, int fieldValue) {
        struct.set(y$constants.LAYOUT, y$constants.OFFSET, fieldValue);
    }

    /**
     * Obtains a slice of {@code arrayParam} which selects the array element at {@code index}.
     * The returned segment has address {@code arrayParam.address() + index * layout().byteSize()}
     */
    public static MemorySegment asSlice(MemorySegment array, long index) {
        return array.asSlice(layout().byteSize() * index);
    }

    /**
     * The size (in bytes) of this struct
     */
    public static long sizeof() { return layout().byteSize(); }

    /**
     * Allocate a segment of size {@code layout().byteSize()} using {@code allocator}
     */
    public static MemorySegment allocate(SegmentAllocator allocator) {
        return allocator.allocate(layout());
    }

    /**
     * Allocate an array of size {@code elementCount} using {@code allocator}.
     * The returned segment has size {@code elementCount * layout().byteSize()}.
     */
    public static MemorySegment allocateArray(long elementCount, SegmentAllocator allocator) {
        return allocator.allocate(MemoryLayout.sequenceLayout(elementCount, layout()));
    }

    /**
     * Reinterprets {@code addr} using target {@code arena} and {@code cleanupAction) (if any).
     * The returned segment has size {@code layout().byteSize()}
     */
    public static MemorySegment reinterpret(MemorySegment addr, Arena arena, Consumer<MemorySegment> cleanup) {
        return reinterpret(addr, 1, arena, cleanup);
    }

    /**
     * Reinterprets {@code addr} using target {@code arena} and {@code cleanupAction) (if any).
     * The returned segment has size {@code elementCount * layout().byteSize()}
     */
    public static MemorySegment reinterpret(MemorySegment addr, long elementCount, Arena arena, Consumer<MemorySegment> cleanup) {
        return addr.reinterpret(layout().byteSize() * elementCount, arena, cleanup);
    }
}
```

Maybe this is too much - however I like how tidy the constants are under the dedicated constant class.

mcimadamore: The above approach can be dialled down in a number of ways:
* maybe a `jextract` option can make all these holder classes "private", so that they don't show up in the API
* for structs, we might only generate holder classes for array fields (on the basis that other info is quickly obtainable when looking at the struct layout).

Stepping back - I think what we need to ask ourselves is whether this incremental complexity of the generated code is worth the added benefits (e.g. of providing more direct access to constants that might be required from the generated code). It might also be useful to take some footprint measurements before/after. It might well be that the extra holder class doesn't matter much in terms of overall bytes of the compiled code. While it's true that there's more classes to load, it's also true that this approach makes everything more lazy - e.g. so that an offset, or a var handle is only computed if the corresponding field is accessed by a client. The two things might well cancel each other out.

JornVernee: This looks like a good approach to me. I think we mostly care about the sources (esp. now that we always generate sources). The cost in terms of size doesn't seem bad. I'm assuming this is for an unfiltered extraction of libpython. I think our assumption is that most of the time, users that care about size will apply filters as well.

As far as source code itself goes. This doesn't look bad. TBH, I think it visually looks a bit cleaner since the constants are 'bagged up' nicely in a separate class that doesn't take up too much space. (at least at the moment. I think we need javadoc for the constant classes as well).

Considering discoverability, lets say someone wants to access the `x` field. When typing 'x', auto-complete would show 2 overloads of `x`, and then `x$constants`. That doesn't seem too cluttered, and I think it's clear from the name that the constant class is not for the common 'I just want to read `x`' use case, given the `$` in the name. (I think it's actually helping us out here).

mcimadamore:
Some more numbers for windows.h (unfiltered)

Size of generated sources
vanilla: 57M
holder: 59M

Size of generated classes
vanilla: 85M
holder: 152M

Number of compiled classes:
vanilla: 20562
holder: 35443

I think these results are not as great as the python experiment suggested. Here we have ~2x jump in the number of generated classes/size of compiled artifacts. The size of source files doesn't change much, given that we're mostly just moving declarations around.

Overall, I'm not comfortable with this jump: in the case of global variables and functions, the additional holder class allows access to information that was previously unavailable (method handle, function descriptor, global var segment and layout). This stuff can be used in various ways (e.g. method handle combination, or access a global variable using atomic access primitives).

In the case of plain struct fields the payoff seems a lot less; for arrays we can access to the handle used by jextract, and the dimensions, so that's nice. For plain non-array fields, their offset and layout can easily be discovered from the struct layout. In both cases, the information stored in the constant classes can be derived from other info that has been generated. Granted, deriving the dimensions of an array field is more difficult (probably due also to some deficiencies of the sequence layout API). But the info is there _somehow_ (unlike e.g. the global variable segment/layout, or the function descriptor of a native function, which is currently just hidden).

For this reason, I'm having an hard time justifying such a footprint increase for something that doesn't seem essential (in the sense that client code can get at the same info, perhaps less conveniently, in other ways). For these reasons, I think I'm leaning towards keep the changes in this PR as is (maybe fix the test to discover the array size dynamically, using the layout), and reassess this problem later, based on feedback (and maybe after some future targeted layout API improvements).

mcimadamore:
I've uploaded a new iteration. The main idea is to just add some additional accessors for every function, global variable and struct field. These accessors contain make use of the `$` sign, and provide access to constants that might be useful from the generated code. E.g. `foo$descriptor` is an accessor that returns the function descriptor for `foo`. This way we don't need to introduce additional classes in the generated API. Here are some numbers on windows.h with this new approach:

Size of generated sources
vanilla: 57M
holder: 68M

Size of generated classes
vanilla: 85M
holder: 87M

Number of compiled classes:
vanilla: 20562
holder: 19671

So, the size in generated sources goes up (because of the extra accessors and javadoc). The size of generated classes stays roughly the same, and the number of generated classes actually goes down (as this new strategy is a bit smarter, in that it uses a single holder class per global variable, instead of two as before).

I'm a bit on the fence as to whether we should keep constant accessors on by default, or whether we should provide a flag to enable/disable constant accessors. Any thoughts?

mcimadamore:
For the records, here's what we generate for the usual `Point` class in the new strategy:

```
/**
 * {@snippet lang=c :
 * struct Point {
 *     int x;
 *     int y;
 * }
 * }
 */
public class Point {

    Point() {
        // Should not be called directly
    }

    private static final GroupLayout $LAYOUT = MemoryLayout.structLayout(
        foo_h.C_INT.withName("x"),
        foo_h.C_INT.withName("y")
    ).withName("Point");

    /**
     * The layout of this struct
     */
    public static final GroupLayout layout() {
        return $LAYOUT;
    }

    private static final OfInt x$LAYOUT = (OfInt)$LAYOUT.select(groupElement("x"));

    /**
     * Layout for field:
     * {@snippet lang=c :
     * int x
     * }
     */
    public static final OfInt x$layout() {
        return x$LAYOUT;
    }
    private static final long x$OFFSET = 0;

    /**
     * Offset for field:
     * {@snippet lang=c :
     * int x
     * }
     */
    public static final long x$offset() {
        return x$OFFSET;
    }

    /**
     * Getter for field:
     * {@snippet lang=c :
     * int x
     * }
     */
    public static int x(MemorySegment struct) {
        return struct.get(x$LAYOUT, x$OFFSET);
    }

    /**
     * Setter for field:
     * {@snippet lang=c :
     * int x
     * }
     */
    public static void x(MemorySegment struct, int fieldValue) {
        struct.set(x$LAYOUT, x$OFFSET, fieldValue);
    }

    private static final OfInt y$LAYOUT = (OfInt)$LAYOUT.select(groupElement("y"));

    /**
     * Layout for field:
     * {@snippet lang=c :
     * int y
     * }
     */
    public static final OfInt y$layout() {
        return y$LAYOUT;
    }
    private static final long y$OFFSET = 4;

    /**
     * Offset for field:
     * {@snippet lang=c :
     * int y
     * }
     */
    public static final long y$offset() {
        return y$OFFSET;
    }

    /**
     * Getter for field:
     * {@snippet lang=c :
     * int y
     * }
     */
    public static int y(MemorySegment struct) {
        return struct.get(y$LAYOUT, y$OFFSET);
    }

    /**
     * Setter for field:
     * {@snippet lang=c :
     * int y
     * }
     */
    public static void y(MemorySegment struct, int fieldValue) {
        struct.set(y$LAYOUT, y$OFFSET, fieldValue);
    }

    /**
     * Obtains a slice of {@code arrayParam} which selects the array element at {@code index}.
     * The returned segment has address {@code arrayParam.address() + index * layout().byteSize()}
     */
    public static MemorySegment asSlice(MemorySegment array, long index) {
        return array.asSlice(layout().byteSize() * index);
    }

    /**
     * The size (in bytes) of this struct
     */
    public static long sizeof() { return layout().byteSize(); }

    /**
     * Allocate a segment of size {@code layout().byteSize()} using {@code allocator}
     */
    public static MemorySegment allocate(SegmentAllocator allocator) {
        return allocator.allocate(layout());
    }

    /**
     * Allocate an array of size {@code elementCount} using {@code allocator}.
     * The returned segment has size {@code elementCount * layout().byteSize()}.
     */
    public static MemorySegment allocateArray(long elementCount, SegmentAllocator allocator) {
        return allocator.allocate(MemoryLayout.sequenceLayout(elementCount, layout()));
    }

    /**
     * Reinterprets {@code addr} using target {@code arena} and {@code cleanupAction) (if any).
     * The returned segment has size {@code layout().byteSize()}
     */
    public static MemorySegment reinterpret(MemorySegment addr, Arena arena, Consumer<MemorySegment> cleanup) {
        return reinterpret(addr, 1, arena, cleanup);
    }

    /**
     * Reinterprets {@code addr} using target {@code arena} and {@code cleanupAction) (if any).
     * The returned segment has size {@code elementCount * layout().byteSize()}
     */
    public static MemorySegment reinterpret(MemorySegment addr, long elementCount, Arena arena, Consumer<MemorySegment> cleanup) {
        return addr.reinterpret(layout().byteSize() * elementCount, arena, cleanup);
    }
}
```

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -309,87 +324,169 @@ private static MemoryLayout variadicLayout(Class<?> c) {
...
+        }
+        incrAlign();
+        appendBlankLine();
+        emitDocComment(var, "Layout for variable:");
+        appendLines(STR."""
+                public static \{layoutType} \{javaName}$layout() {
+                    return \{mangledName}.LAYOUT;
+                }
+                """);
+        if (!Utils.isStructOrUnion(varType) && !Utils.isArray(varType)) {
```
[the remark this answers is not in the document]
mcimadamore: Because for structs  and arrays, the getter is already returning the global variable segment

src/main/java/org/openjdk/jextract/impl/StructBuilder.java
```diff
@@ -133,13 +134,21 @@ public void addFunctionalInterface(Declaration parentDecl, Type.Function funcTyp
...
         appendBlankLine();
-        String offsetField = emitOffsetFieldDecl(varTree);
+        String layoutField = emitLayoutFieldDecl(varTree, javaName);
+        appendBlankLine();
+        String offsetField = emitOffsetFieldDecl(varTree, javaName);
         if (Utils.isArray(varTree.type()) || Utils.isStructOrUnion(varTree.type())) {
-            String sizeField = emitSizeFieldDecl(varTree);
-            emitSegmentGetter(javaName, varTree, offsetField, sizeField);
-            emitSegmentSetter(javaName, varTree, offsetField, sizeField);
+            emitDimensionsFieldDecl(varTree, javaName);
```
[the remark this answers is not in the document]
mcimadamore: this seems a mistake

JornVernee (approved):

PR #201, opened 2024-02-07
7903654: Jextract should use library lookup

mcimadamore:
Historically, jextract has always used `System::loadLibrary` and `System::load` to load libraries specified via the `-l` option. This is suboptimal, as in the most common case, such libraries will be installed using e.g. `apt`, in which case `System::load/loadLibrary` is of little use, as it cannot find libraries in path known to the dynamic linker.

This patch changes the default for library loading to use `SymbolLookup::libraryLookup` instead. That is, in absence of any other options, each library provided via `-l` will create a new `SymbolLookup::loaderLookup`, associated with the library arena.

The `-l` option also supports a new disambiguation mechanism, borrowed from the gcc linker options - that is, if the library name starts with `:`, what follows is parsed as a path. Otherwise, we assume it's a library name, and we add prefix/suffix, using `System::mapLibraryName` before creating the lookup.

In cases where the old behavior is preferred (e.g. if libraries to be loaded sit in `java.library.path`) a fallback option has been provided, namely `--use-system-load-library`.

Regardless of the options used, the header class will feature a composite lookup. Inside this lookup we find:

1. one or more library lookup (one for each `-l` option)
2. the loader lookup
3. the native linker's default lookup

When `--use-system-load-library` is used, (1) is empty, and a static initializer block with the corresponding calls to `System::load`, `System::loadLibrary` is added instead.

---------
### Progress

### Issue
 * [CODETOOLS-7903654](https://bugs.openjdk.org/browse/CODETOOLS-7903654): Jextract should use library lookup (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/201/head:pull/201` \
`$ git checkout pull/201`

`$ git checkout pull/201` \
`$ git pull https://git.openjdk.org/jextract.git pull/201/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 201`

`$ git pr show -t 201`

</details>

<a href="https://git.openjdk.org/jextract/pull/201.diff">https://git.openjdk.org/jextract/pull/201.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/201#issuecomment-1933009851)


test/jtreg/generator/allocCallback/TestAllocCallback.java
```diff
@@ -33,7 +33,7 @@
  * @bug 7903239
  * @summary ofAddress factory of function pointer type is wrong for struct returns
  * @library /lib
- * @run main/othervm JtregJextract -l AllocCallback -t test.jextract.allocCallback alloc_callback.h
+ * @run main/othervm JtregJextract -l AllocCallback --use-system-load-library -t test.jextract.allocCallback alloc_callback.h
```
mcimadamore, line 36: In this (and other) test, we use the `--use-system-load-library` as the build will put the shared libraries in the `java.library.path`, and pass that variable to jtreg.

test/lib/JtregJextract.java
```diff
@@ -71,16 +71,6 @@ private String[] processArgs(String... args) {
                 break;
             }
 
-            if ("-libpath".equals(opt)) {
```
mcimadamore, line 74: I've removed this, as I think this was a questionable option. I'm not exactly sure what problems this was meant to address, but the only test using this has been rewritten to use other options instead.

JornVernee (approved):
Looks good.

I suppose it's hard to test this using the current testing infrastructure though. I gave it a try here with the libclang sample, and it works as expected.

PR #203, opened 2024-02-09
7903656: Variadic invoker class names are not mangled

mcimadamore:
This PR fixes an issue with the variadic invoker nested class names that are not mangled correctly. This might create issues for case-insensitive file systems.

While working on this, I noticed also that the current test we have for variadic classes (TestPrintf) was not executing as expected, because of a typo in a data provider. After fixing the test code, the test started to fail.

One of the test cases was expecting an `IllegalArgumentException` when passing an int value to a double variadic parameter. But, as we use a method handle argument spreader, no exception is throw, and the value is instead silently converted. I tweaked the test a bit, to create a situation where no conversion is possible, and I noticed that the call failed with `ClassCastException`, but that exception is not handled by the variadic invoker code, so I fixed that as well.

---------
### Progress

### Issue
 * [CODETOOLS-7903656](https://bugs.openjdk.org/browse/CODETOOLS-7903656): Variadic invoker class names are not mangled (**Bug** - P3)

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to [5080d83c](https://git.openjdk.org/jextract/pull/203/files/5080d83caaa50b9ba8ed9ee1f81351084c7d1d69)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/203/head:pull/203` \
`$ git checkout pull/203`

`$ git checkout pull/203` \
`$ git pull https://git.openjdk.org/jextract.git pull/203/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 203`

`$ git pr show -t 203`

</details>

<a href="https://git.openjdk.org/jextract/pull/203.diff">https://git.openjdk.org/jextract/pull/203.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/203#issuecomment-1935548576)

### DIFF
--- a/src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
+++ b/src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
@@ -220,7 +220,7 @@ class HeaderFileBuilder extends ClassSourceBuilder {
                                 traceDowncall(\{traceArgList});
                             }
                             \{returnWithCast}mh$.invokeExact(\{paramList});
-                        } catch(IllegalArgumentException ex$)  {
+                        } catch(IllegalArgumentException | ClassCastException ex$)  {
                             throw ex$; // rethrow IAE from passing wrong number/type of args
                         } catch (Throwable ex$) {
                            throw new AssertionError("should not reach here", ex$);
--- a/src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
+++ b/src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
@@ -190,12 +190,13 @@ class HeaderFileBuilder extends ClassSourceBuilder {
             }
             """);
         } else {
+            String invokerClassName = newHolderClassName(javaName);
             String paramExprs = paramExprs(declType, finalParamNames, isVarArg);
             String varargsParam = finalParamNames.get(finalParamNames.size() - 1);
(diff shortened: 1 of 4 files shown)


src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -589,7 +590,7 @@ private void emitPrimitiveTypedefLayout(String javaName, Type type, Declaration
     }
 
     private String newHolderClassName(String javaName) {
-        String holderClassName = STR."\{javaName}$constants";
+        String holderClassName = javaName;
```
mcimadamore, line 593: Strictly speaking, this change is not necessary - but since we emit nested classes with same name as other members for variadic functions too, I don't really see a reason to append the `$constant` prefix to the nested constant class here.

mcimadamore: In the end, I've reverted this - I realized that I didn't "clean" before running tests and there were several conflicts of the kind already described. The problem is that it is sometimes common to do this:

```
struct foo { ... } foo;
```

Which then leads to ambiguities. So I reverted that back. (for variadic, there's no such issue).

PR #204, opened 2024-02-09
Add missing inputs to tasks in gradle build

JornVernee:
I recently re-did some of the task dependencies of the gradle build: https://github.com/openjdk/jextract/pull/200

From my initial testing for that PR, and reading online, I was under the impression that tasks would re-run when the outputs of a dependency changes. After additional use & testing, this turns out to be not the case and we need to explicitly set task inputs any way. This PR does that. 

I also changed the `copyLibClang` task from `Copy` to `Sync`. If a file is removed from the source directory, `Copy` will not remove it from the output, but `Sync` will. This helps ensure that we don't package any stale files into the jextract jmod (which might create include conflicts). This is also something I noticed during more thorough testing.

Finally, I noticed that on Windows we were copying `clang.exe` to the jextract jmod archive, since we copy from the `bin` dir on Windows, which contains both the `libclang.dll` and the `clang.exe` file. I've added an exclude for `clang.exe`.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/204/head:pull/204` \
`$ git checkout pull/204`

`$ git checkout pull/204` \
`$ git pull https://git.openjdk.org/jextract.git pull/204/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 204`

`$ git pr show -t 204`

</details>

<a href="https://git.openjdk.org/jextract/pull/204.diff">https://git.openjdk.org/jextract/pull/204.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/204#issuecomment-1936092781)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -65,7 +65,7 @@ jar {
     archiveVersion = project.version
 }
 
-task copyLibClang(type: Copy) {
+task copyLibClang(type: Sync) {
     into("$buildDir/jmod_inputs")
 
     from("${libclang_dir}") {
@@ -84,6 +84,8 @@ task createJextractJmod(type: Exec) {
     dependsOn jar, copyLibClang
 
     // if these inputs or outputs change, gradle will rerun the task
+    inputs.file(jar.archiveFile.get())
+    inputs.dir("$jextract_jmod_inputs")
     outputs.file(jextract_jmod_file)
 
     doFirst {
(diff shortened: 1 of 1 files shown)


mcimadamore (approved):
I wonder if this is why I didn't see the build failure on incremental rebuild when working onmy previous PR. Thanks for fixing.

PR #214, opened 2024-02-16
7903666: Jextract should report issues with missing dependencies

mcimadamore:
When custom include filters are specified on the command line, it is possible for jextract to generate non-sensical code (now even more so than in the past). Possible issues:

* a function descriptor refers to a non-existent struct/union
* a struct/union typedef refers to a non-existent struct/union
* a struct/union field refers to a non-existent struct/union
* a global variable refers to a non-existent struct/union

In such cases, jextract should terminate (and report the problem to the user), rather than generating buggy code.

The fix is relatively simple, and amounts at checking whether a declaration depends on a skipped struct in `IncludeHelper`. That said, when we detect an issue, we should stop jextract. So we need some way to notify jextract that some error occurred, and stop.

I initially had a hacky approach with some mutable field in `IncludeFilter`, but I've decided to implement a minimal Logger class, and make _all_ error/warnings go through the Logger. This way we can easily check how many errors occurred, and stop jextract if needed. Of course, this led to several changes, as I realized that we were being inconsistent on our usage of the property file. I now made sure that all messages use some key in the property file (e.g. no lone strings). This might be important if jextract message will ever be translated to languages other than English.

---------
### Progress

### Issue
 * [CODETOOLS-7903666](https://bugs.openjdk.org/browse/CODETOOLS-7903666): Jextract should report issues with missing dependencies (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/214/head:pull/214` \
`$ git checkout pull/214`

`$ git checkout pull/214` \
`$ git pull https://git.openjdk.org/jextract.git pull/214/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 214`

`$ git pr show -t 214`

</details>

<a href="https://git.openjdk.org/jextract/pull/214.diff">https://git.openjdk.org/jextract/pull/214.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/214#issuecomment-1948663270)

### DIFF
--- a/src/main/java/org/openjdk/jextract/JextractTool.java
+++ b/src/main/java/org/openjdk/jextract/JextractTool.java
@@ -31,6 +31,7 @@ import org.openjdk.jextract.impl.CommandLine;
 import org.openjdk.jextract.impl.DuplicateFilter;
 import org.openjdk.jextract.impl.IncludeFilter;
 import org.openjdk.jextract.impl.IncludeHelper;
+import org.openjdk.jextract.impl.Logger;
 import org.openjdk.jextract.impl.NameMangler;
 import org.openjdk.jextract.impl.Options.Library;
 import org.openjdk.jextract.impl.OutputFactory;
@@ -65,15 +66,8 @@ import java.util.stream.Stream;
  * on top of the underlying memory access var handles. For each struct, a static layout field is generated.
  */
 public final class JextractTool {
-    private static final String MESSAGES_RESOURCE = "org.openjdk.jextract.impl.resources.Messages";
-
-    private static final ResourceBundle MESSAGES_BUNDLE;
-    static {
-        MESSAGES_BUNDLE = ResourceBundle.getBundle(MESSAGES_RESOURCE, Locale.getDefault());
-    }
 
     public static final boolean DEBUG = Boolean.getBoolean("jextract.debug");
-    public static final Optional<Path> PLATFORM_INCLUDE_PATH = inferPlatformIncludePath();
 
     // error codes
     private static final int SUCCESS       = 0;
@@ -83,16 +77,10 @@ public final class JextractTool {
     private static final int RUNTIME_ERROR = 4;
     private static final int OUTPUT_ERROR  = 5;
 
-    private final PrintWriter out;
-    private final PrintWriter err;
-
-    private static String format(String msgId, Object... args) {
-        return new MessageFormat(MESSAGES_BUNDLE.getString(msgId)).format(args);
-    }
+    private final Logger logger;
 
     private JextractTool(PrintWriter out, PrintWriter err) {
-        this.out = out;
-        this.err = err;
+        this.logger = new Logger(out, err);
     }
 
     private static Path generateTmpSource(List<Path> headers) {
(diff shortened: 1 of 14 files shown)


src/main/java/org/openjdk/jextract/impl/IncludeFilter.java
```diff
@@ -93,11 +100,26 @@ public Void visitVariable(Declaration.Variable tree, Declaration parent) {
...
+        warnMissingDep(parent != null ? parent : tree, tree.type());
         return null;
     }
 
     @Override
     public Void visitDeclaration(Declaration decl, Declaration parent) {
         return null;
     }
+
+    void warnMissingDep(Declaration decl, Type type) {
```
mcimadamore, line 112: This is the meat of the fix

src/main/java/org/openjdk/jextract/impl/OutputFactory.java
```diff
@@ -64,11 +64,7 @@ JavaSourceFile[] generate(Declaration.Scoped decl) {
     }
 
     private void generateDecl(Declaration tree) {
-        try {
-            tree.accept(this, null);
-        } catch (Exception ex) {
-            ex.printStackTrace();
-        }
+        tree.accept(this, null);
```
mcimadamore, line 67: the try/catch seemed suspicious, and was also printing on std err, so I dropped that (after all, if some unexpected exception happens, we want to know?)

src/main/java/org/openjdk/jextract/JextractTool.java
```diff
@@ -115,34 +103,34 @@ private static Path generateTmpSource(List<Path> headers) {
...
+                .map(new IncludeFilter(includeHelper, logger)::scan)
                 .map(new DuplicateFilter()::scan)
                 .map(new NameMangler(headerName)::scan)
-                .map(new UnsupportedFilter(errStream)::scan)
+                .map(new UnsupportedFilter(logger)::scan)
                 .findFirst().get();
-        return OutputFactory.generateWrapped(transformedDecl, targetPkg, libs, useSystemLoadLibrary);
+        return logger.hasErrors() ?
+                List.of() :
+                List.of(OutputFactory.generateWrapped(transformedDecl, targetPkg, libs, useSystemLoadLibrary));
```
JornVernee, line 133: So, we now don't generate anything when there are errors.

I think we should also return a non-zero exit code in that case.

src/main/java/org/openjdk/jextract/impl/OutputFactory.java
JornVernee: I suppose this was added when jextract was less stable, or lacked filtering support. So, if we got an exception while generating one of the trees, jextract would still output _something_

test/testng/org/openjdk/jextract/test/toolprovider/includeDeps/TestBadIncludes.java
```diff
@@ -0,0 +1,63 @@
...
+        Path output = getOutputFilePath("TestBadIncludes-badIncludes.h");
+        Path outputH = getInputFilePath("bad_includes.h");
+        result = run(output,
+        "--include-struct", "B",
+                "--include-function", "m",
+                "--include-typedef", "T",
+                "--include-struct", "C",
+                "--include-function", "n",
+                outputH.toString());
+    }
```
JornVernee, line 47: Maybe `result.checkFailure()` here too?

JornVernee (approved):

test/testng/org/openjdk/jextract/test/toolprovider/includeDeps/TestBadIncludes.java
mcimadamore: whoops - yes!

src/main/java/org/openjdk/jextract/JextractTool.java
mcimadamore: True

JornVernee (approved):

PR #217, opened 2024-02-17
7903672: Some issues with missing dependency errors

JornVernee:
- Only error on missing dependencies, if the dependent is not skipped
- Scan element types of arrays for missing dependencies as well
- Don't issue warnings about unsupported types/fields for skipped elements

I've pulled the code that issues errors for missing dependencies into a separate phase (`MissingDepWarner`), since the control flow in some of the visitor methods was kinda tricky to follow if this stayed mixed in with the `IncludeFilter` code. The new phase just cleanly checks whether the visited element has a `Skip` on it and then returns at the start, before doing any work.

I've also re-ordered the phases a bit so that the phases that add Skips run first, so that we don't issue warnings/errors for things that are skipped any way by a later phase. Since `UnsupportedFilter` both adds Skips and issues warnings, it sits right in the middle.

Q: should we also return early in `NameMangler` if an element is skipped? (Right now it's just doing extra work).

---------
### Progress

### Issue
 * [CODETOOLS-7903672](https://bugs.openjdk.org/browse/CODETOOLS-7903672): Some issues with missing dependency errors (**Bug** - P4)

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - **Reviewer**) ⚠️ Review applies to [fa6e182b](https://git.openjdk.org/jextract/pull/217/files/fa6e182b5a31027cdedf695791f612c5e027f861)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/217/head:pull/217` \
`$ git checkout pull/217`

`$ git checkout pull/217` \
`$ git pull https://git.openjdk.org/jextract.git pull/217/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 217`

`$ git pr show -t 217`

</details>

<a href="https://git.openjdk.org/jextract/pull/217.diff">https://git.openjdk.org/jextract/pull/217.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/217#issuecomment-1952585830)

### DIFF
--- a/src/main/java/org/openjdk/jextract/impl/MissingDepWarner.java
+++ b/src/main/java/org/openjdk/jextract/impl/MissingDepWarner.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2024 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2024 Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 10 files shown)


src/main/java/org/openjdk/jextract/impl/MissingDepWarner.java
```diff
@@ -0,0 +1,104 @@
...
+
+import org.openjdk.jextract.Declaration;
+import org.openjdk.jextract.Type;
+import org.openjdk.jextract.Type.Delegated;
+import org.openjdk.jextract.impl.DeclarationImpl.Skip;
+
+/*
+ * This visitor marks declarations to be skipped, based on --include options specified.
+ */
+public final class MissingDepWarner implements Declaration.Visitor<Void, Declaration> {
```
mcimadamore, line 35: We should probably call this `MissingDepChecker` - given we give errors, not warnings (similar comment applies to the `warnMissingDeps` method)

PR #218, opened 2024-02-19
7903670: jextract generates wrong packed struct layout

mcimadamore:
The change to refer to other struct layouts by name resulted in an interesting issue with the `pragma pack` support. When jextract simply re-inlined all nested struct layouts into the enclosing struct, it was simply necessary to add as many calls to `withByteAlignment` as needed, to force struct alignment to adhere to the pragma directive.

But now that nested layouts just refer to a layout declared somewhere else, we can no longer add `withByteAlignment` to the fields of the _nested_ struct. So we run into issues if:

* the enclosing struct is packed, and
* the nested struct is declared as non-packed

Because, in this case, the layout declaration of the nested struct just contains natural alignment constraints. As such, when this layout is referred to by the enclosing struct layout (e.g. in a field), packed alignment constraints are violated. Note also that simply calling `withByteAlignment` on the nested struct is not enough, as the struct alignment can be stricter than the alignment of any of the fields.

In other words, we'd need to duplicate the nested layout into the packed enclosing struct, and decorate all its fields with the necessary `withByteAlignment` calls. I found this solution not satisfying, as now we'd have _two_ (or more!) different declaration for the layout of a struct: one corresponding to the place where it was declared; another for the place where it was _used_ with stricter alignment constraints. This is bad, because if users tweak the layout of the nested struct in any way (e.g. using big endian order for some of its fields), such changes will _not_ be reflected by the packed layout view.

For this reason, I've opted for a solution where packed views are generated _dynamically_, with the help of a compact function in the main header class. This allows jextract to pack layouts (of any kind) at will, according to the constraints of the struct that is currently emitting.

In other words, given this:

```c
typedef union epoll_data
{
  void *ptr;
  int fd;
  int u32;
  long u64;
} epoll_data_t;

#pragma pack(1)
struct epoll_event
{
  int events;
  epoll_data_t data;
} __EPOLL_PACKED; 
```

We now generate:

```java
private static final GroupLayout $LAYOUT = MemoryLayout.unionLayout(
        foo_h.C_POINTER.withName("ptr"),
        foo_h.C_INT.withName("fd"),
        foo_h.C_INT.withName("u32"),
        foo_h.C_LONG.withName("u64")
    ).withName("epoll_data");

private static final GroupLayout $LAYOUT = MemoryLayout.structLayout(
        foo_h.align(foo_h.C_INT, 1).withName("events"),
        foo_h.align(epoll_data.layout(), 1).withName("data")
    ).withName("epoll_event");
```

Note the calls to `foo_h::align` in the `epoll_event` packed layout.

---------
### Progress

### Issue
 * [CODETOOLS-7903670](https://bugs.openjdk.org/browse/CODETOOLS-7903670): jextract generates wrong packed struct layout (**Bug** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/218/head:pull/218` \
`$ git checkout pull/218`

`$ git checkout pull/218` \
`$ git pull https://git.openjdk.org/jextract.git pull/218/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 218`

`$ git pr show -t 218`

</details>

<a href="https://git.openjdk.org/jextract/pull/218.diff">https://git.openjdk.org/jextract/pull/218.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/218#issuecomment-1952220394)

### DIFF
--- a/src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
+++ b/src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
@@ -32,6 +32,7 @@ import org.openjdk.jextract.Type.Declared;
 import org.openjdk.jextract.Type.Delegated;
 import org.openjdk.jextract.Type.Function;
 import org.openjdk.jextract.Type.Primitive;
+import org.openjdk.jextract.impl.DeclarationImpl.ClangAlignOf;
 import org.openjdk.jextract.impl.DeclarationImpl.DeclarationString;
 import org.openjdk.jextract.impl.DeclarationImpl.JavaName;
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
```diff
@@ -191,10 +192,10 @@ String layoutString(Type type, long align) {
         return switch (type) {
             case Primitive p -> primitiveLayoutString(p, align);
             case Declared d when Utils.isEnum(d) -> layoutString(((Constant)d.tree().members().get(0)).type(), align);
-            case Declared d when Utils.isStructOrUnion(d) -> STR."\{JavaName.getFullNameOrThrow(d.tree())}.layout()";
-            case Delegated d when d.kind() == Delegated.Kind.POINTER -> STR."\{runtimeHelperName()}.C_POINTER";
+            case Declared d when Utils.isStructOrUnion(d) -> alignIfNeeded(STR."\{JavaName.getFullNameOrThrow(d.tree())}.layout()", ClangAlignOf.getOrThrow(d.tree()) / 8, align);
```
mcimadamore, line 195: Here, I also realized that:
* address layouts were not being aligned
* same for function pointers
* and same for arrays
(I added a test for all these cases)

JornVernee (approved):

PR #219, opened 2024-02-19
7903673: Jextract throws NPE when generating enum constant javadoc

mcimadamore:
This PR fixes `TreeMaker` so that it doesn't try to generate javadoc for `null` enum constants.

Unfortunately I was not able to reproduce on Linux. The problem seems to appear on MacOS when extracting this enum:

```
enum __CFByteOrder {
    CFByteOrderUnknown,
    CFByteOrderLittleEndian,
    CFByteOrderBigEndian
}; 
```

But there doesn't seem to be anything peculiar with this (and this does work when extracted in isolation).
It might indicate a deeper issue in libclang. For now, I made the code more robust, so that it handles possible null enum constants (these are left out from the IR).

---------
### Progress

### Issue
 * [CODETOOLS-7903673](https://bugs.openjdk.org/browse/CODETOOLS-7903673): Jextract throws NPE when generating enum constant javadoc (**Bug** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/219/head:pull/219` \
`$ git checkout pull/219`

`$ git checkout pull/219` \
`$ git pull https://git.openjdk.org/jextract.git pull/219/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 219`

`$ git pr show -t 219`

</details>

<a href="https://git.openjdk.org/jextract/pull/219.diff">https://git.openjdk.org/jextract/pull/219.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/219#issuecomment-1952518551)

### DIFF
--- a/src/main/java/org/openjdk/jextract/impl/TreeMaker.java
+++ b/src/main/java/org/openjdk/jextract/impl/TreeMaker.java
@@ -370,7 +370,7 @@ class TreeMaker {
             List<Declaration> decls = new ArrayList<>();
             c.forEach(child -> {
                 Declaration enumConstantDecl = createTree(child);
-                if (enumConstantDecl != null) {
+                if (enumConstantDecl != null) { // see CODETOOLS-7903673
                     DeclarationString.with(enumConstantDecl, enumConstantString(c.spelling(), (Declaration.Constant) enumConstantDecl));
                     decls.add(enumConstantDecl);
                 }
--- a/src/main/java/org/openjdk/jextract/impl/TreeMaker.java
+++ b/src/main/java/org/openjdk/jextract/impl/TreeMaker.java
@@ -366,13 +366,14 @@ class TreeMaker {
     }
 
     public Declaration.Scoped createEnum(Cursor c) {
-        List<Declaration> decls = new ArrayList<>();
-        c.forEach(child -> decls.add(createTree(child)));
         if (c.isDefinition()) {
-            //just a declaration AND definition, we have a layout
-            decls.forEach(d -> {
-                // append declaration string
-                DeclarationString.with(d, enumConstantString(c.spelling(), (Declaration.Constant)d));
+            List<Declaration> decls = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


mcimadamore:
A possible explanation as to why we're seeing this now is that we now generate enum constant comments much earlier. Perhaps this was never an issue before as this enum was skipped.

mcimadamore:
Also, our `makeTreeInternal` checks that a position is set:

```
if (pos == Position.NO_POSITION) return null; // intrinsic, skip
```

This code is also new, and could lead to this issue.

src/main/java/org/openjdk/jextract/impl/TreeMaker.java
```diff
@@ -366,13 +366,14 @@ private static OptionalLong offsetOfAnonymousRecord(Cursor outermostParent, Curs
...
-        c.forEach(child -> decls.add(createTree(child)));
         if (c.isDefinition()) {
-            //just a declaration AND definition, we have a layout
-            decls.forEach(d -> {
-                // append declaration string
-                DeclarationString.with(d, enumConstantString(c.spelling(), (Declaration.Constant)d));
+            List<Declaration> decls = new ArrayList<>();
+            c.forEach(child -> {
+                Declaration enumConstantDecl = createTree(child);
+                if (enumConstantDecl != null) {
```
JornVernee, line 373: Could you add a comment on this `if` that points back at the bug? (just in case we ever wonder why there is a `null` check here in the future)

JornVernee (approved):

JornVernee (approved):

sundararajana:
I was able to reproduce NPE on MacOS / x64 without patch. With the current patch, I tried extracting user headers on Mac OS/x64. No more NPE.

sundararajana (approved):
LGTM

PR #221, opened 2024-02-20
7903674: jextract should ignore non-enum constants inside enum

sundararajana:
* checking cursor kind of child cursor to be CursorKind.EnumConstantDecl. 
* Piggybacking to fix similar check of record child element kinds to be CursorKind.FieldDecl.  
* Added null check for createTree return value in collectNestedTypes method as well.
* In addition to jextract tests, I ran all samples on Mac OS. All fine with this change. Also, ran jextract script for Mac OS from  https://github.com/manuelbl/JavaDoesUSB.git.

---------
### Progress

### Issue
 * [CODETOOLS-7903674](https://bugs.openjdk.org/browse/CODETOOLS-7903674): jextract should ignore non-enum constants inside enum (**Bug** - P3)

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - **Reviewer**) ⚠️ Review applies to [aca001e3](https://git.openjdk.org/jextract/pull/221/files/aca001e3a69cd30575bbc1b2f16a8aaf03452584)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/221/head:pull/221` \
`$ git checkout pull/221`

`$ git checkout pull/221` \
`$ git pull https://git.openjdk.org/jextract.git pull/221/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 221`

`$ git pr show -t 221`

</details>

<a href="https://git.openjdk.org/jextract/pull/221.diff">https://git.openjdk.org/jextract/pull/221.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/221#issuecomment-1953615221)

### DIFF
--- a/src/main/java/org/openjdk/jextract/impl/TreeMaker.java
+++ b/src/main/java/org/openjdk/jextract/impl/TreeMaker.java
@@ -289,15 +289,13 @@ class TreeMaker {
                     if (fc.isAnonymousStruct()) {
                         // process struct recursively
                         pendingFields.add(recordDeclaration(parent, fc).tree());
-                    } else {
-                        if (fc.kind() == CursorKind.FieldDecl) {
-                            Declaration fieldDecl = createTree(fc);
-                            ClangSizeOf.with(fieldDecl, fc.type().kind() == TypeKind.IncompleteArray ?
+                    } else if (fc.kind() == CursorKind.FieldDecl) {
+                        Declaration fieldDecl = createTree(fc);
+                        ClangSizeOf.with(fieldDecl, fc.type().kind() == TypeKind.IncompleteArray ?
                                     0 : fc.type().size() * 8);
-                            ClangOffsetOf.with(fieldDecl, parent.type().getOffsetOf(fc.spelling()));
-                            ClangAlignOf.with(fieldDecl, fc.type().align() * 8);
-                            pendingFields.add(fieldDecl);
-                        }
+                        ClangOffsetOf.with(fieldDecl, parent.type().getOffsetOf(fc.spelling()));
+                        ClangAlignOf.with(fieldDecl, fc.type().align() * 8);
+                        pendingFields.add(fieldDecl);
                     }
                 }
             } else {
--- a/src/main/java/org/openjdk/jextract/impl/TreeMaker.java
+++ b/src/main/java/org/openjdk/jextract/impl/TreeMaker.java
@@ -289,7 +289,7 @@ class TreeMaker {
                     if (fc.isAnonymousStruct()) {
                         // process struct recursively
                         pendingFields.add(recordDeclaration(parent, fc).tree());
-                    } else if (fc.kind() == CursorKind.FieldDecl) {
+                    } else {
                         Declaration fieldDecl = createTree(fc);
                         ClangSizeOf.with(fieldDecl, fc.type().kind() == TypeKind.IncompleteArray ?
                                     0 : fc.type().size() * 8);
@@ -474,9 +474,7 @@ class TreeMaker {
                     collectNestedTypes(m, nestedTypes, ignoreNestedParams);
                 } else {
                     Declaration decl = createTree(m);
-                    if (decl != null) {
-                        nestedTypes.add(decl);
-                    }
+                    nestedTypes.add(decl);
                 }
             }
         });
--- a/src/main/java/org/openjdk/jextract/impl/TreeMaker.java
+++ b/src/main/java/org/openjdk/jextract/impl/TreeMaker.java
@@ -290,12 +290,14 @@ class TreeMaker {
                         // process struct recursively
                         pendingFields.add(recordDeclaration(parent, fc).tree());
                     } else {
-                        Declaration fieldDecl = createTree(fc);
-                        ClangSizeOf.with(fieldDecl, fc.type().kind() == TypeKind.IncompleteArray ?
-                                0 : fc.type().size() * 8);
-                        ClangOffsetOf.with(fieldDecl, parent.type().getOffsetOf(fc.spelling()));
-                        ClangAlignOf.with(fieldDecl, fc.type().align() * 8);
-                        pendingFields.add(fieldDecl);
+                        if (fc.kind() == CursorKind.FieldDecl) {
+                            Declaration fieldDecl = createTree(fc);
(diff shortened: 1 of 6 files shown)


mcimadamore:
Is there some kind of tests we can add for the new logic? I guess I'm still struggling to understand what kind of cursor kind can cause this havoc - the C grammar doesn't seem to support much beside enum constants inside an enum?

sundararajana:
> Is there some kind of tests we can add for the new logic? I guess I'm still struggling to understand what kind of cursor kind can cause this havoc - the C grammar doesn't seem to support much beside enum constants inside an enum?

Attribute kind cursor.

src/main/java/org/openjdk/jextract/impl/TreeMaker.java
```diff
@@ -289,10 +289,10 @@ final Type.Declared recordDeclaration(Cursor parent, Cursor recordCursor) {
                     if (fc.isAnonymousStruct()) {
                         // process struct recursively
                         pendingFields.add(recordDeclaration(parent, fc).tree());
-                    } else {
+                    } else if (fc.kind() == CursorKind.FieldDecl) {
```
mcimadamore, line 292: This extra check seems not very useful. Note that this section is guarded by "isFlattenable" which will exclude attributes, and stuff that we do not support.

src/main/java/org/openjdk/jextract/impl/TreeMaker.java
```diff
@@ -471,7 +473,10 @@ private void collectNestedTypes(Cursor c, List<Declaration> nestedTypes, boolean
                 if (m.kind() == CursorKind.ParmDecl && !ignoreNestedParams) {
                     collectNestedTypes(m, nestedTypes, ignoreNestedParams);
                 } else {
-                    nestedTypes.add(createTree(m));
+                    Declaration decl = createTree(m);
```
mcimadamore, line 476: note that the output of this method is only used by `withNestedTypes` which already filters the output, looking for scoped declarations. E.g. it has `d instanceof Scoped` which will fail if the decl is `null`. So I don't think this adds much, but perhaps it could be slightly cleaner to filter out nulls as you did.

mcimadamore:
I believe that at least test for something like this:

```
enum __attribute((deprecated)) Foo {
    CFByteOrderUnknown,
}; 
```

Should be added. Windows uses a different attribute syntax (`[[deprecated]]`). So either we use a macro to define the attribute, or we just test on Linux/Mac.

It might also be useful to add tests for:

```
struct __attribute((deprecated))  Foo {
   int x;
};
```

e.g. struct field case. And:

```
typedef __attribute((deprecated)) struct  { int x } BAR;
```

For the nested struct case.

src/main/java/org/openjdk/jextract/impl/TreeMaker.java
mcimadamore: Actually, debugging some more I realized that in this case we first check `m.isDefinition()` which always returns false for attributes. So I don't think we can ever get nulls here.

sundararajana:
> I believe that at least test for something like this:
> 
> ```
> enum __attribute((deprecated)) Foo {
>     CFByteOrderUnknown,
> }; 
> ```
> 
> Should be added. Windows uses a different attribute syntax (`[[deprecated]]`). So either we use a macro to define the attribute, or we just test on Linux/Mac.
> 
> It might also be useful to add tests for:
> 
> ```
> struct __attribute((deprecated))  Foo {
>    int x;
> };
> ```
> 
> e.g. struct field case. And:
> 
> ```
> typedef __attribute((deprecated)) struct  { int x } BAR;
> ```
> 
> For the nested struct case.

Will add a test

test/testng/org/openjdk/jextract/test/api/attributes/Test7903674.java
```diff
@@ -0,0 +1,43 @@
...
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package org.openjdk.jextract.test.api.attributes;
+
+import org.testng.annotations.Test;
+import testlib.JextractApiTestBase;
+
+public class Test7903674 extends JextractApiTestBase {
```
JornVernee, line 28: Could you please give a descriptive name to the test class. It becomes very hard to find anything with all the bug numbers.

src/main/java/org/openjdk/jextract/impl/TreeMaker.java
```diff
@@ -369,10 +369,12 @@ public Declaration.Scoped createEnum(Cursor c) {
         if (c.isDefinition()) {
             List<Declaration> decls = new ArrayList<>();
             c.forEach(child -> {
-                Declaration enumConstantDecl = createTree(child);
-                if (enumConstantDecl != null) { // see CODETOOLS-7903673
-                    DeclarationString.with(enumConstantDecl, enumConstantString(c.spelling(), (Declaration.Constant) enumConstantDecl));
-                    decls.add(enumConstantDecl);
+                if (child.kind() == CursorKind.EnumConstantDecl) {
+                    Declaration enumConstantDecl = createTree(child);
+                    if (enumConstantDecl != null) {
```
JornVernee, line 374: The null check here is not needed anymore. (test coverage shows the `null` branch is uncovered here as well)

mcimadamore: Isn't that the whole point of the fix? (e.g. skipping nulls when a "weird" cursor shows up)

mcimadamore (approved):
Looks good to me - please rename the test as @JornVernee suggested

src/main/java/org/openjdk/jextract/impl/TreeMaker.java
JornVernee: I thought the `child.kind() == CursorKind.EnumConstantDecl` supersedes that

mcimadamore: Ah, I missed the new check for EnumConstantDecl. I don't think we can ever return a null in there. So, yes, the extra null check is redundant.

PR #231, opened 2024-04-09
Add jextract guide

JornVernee:
Add a comprehensive jextract guide under a new `doc/` folder.

This is meant as a comprehensive guide about the features of jextract and the code that it generates (including both examples from header files, the code that jextract generates, and corresponding Java user code). This is a first cut, and I'm anticipating quite a bit of comments.

Some sections of the readme have been moved to the guide (with minor edits here and there), and replaced by a link in the readme.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/231/head:pull/231` \
`$ git checkout pull/231`

`$ git checkout pull/231` \
`$ git pull https://git.openjdk.org/jextract.git pull/231/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 231`

`$ git pr show -t 231`

</details>

<a href="https://git.openjdk.org/jextract/pull/231.diff">https://git.openjdk.org/jextract/pull/231.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/231#issuecomment-2045736080)


doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
+# Jextract Guide
+
+The jextract tool parses header (.h) files of native libraries, and generates Java code,
```
nlisker, line 3: Aren't .c files also viable for jextract?

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+called 'bindings', which  use the Foreign Function and Memory API (FFM API) under the hood,
```
nlisker, line 4: Double space after `which`.

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+that can be used by a client to access the native library.
+
+Interacting with native C code through the FFM API works
+by loading a native library (e.g. a `.so`/`.dll`/`.dylib` file), which is essentially an
```
nlisker, line 8: Comma after `e.g.`.

doc/GUIDE.md
JornVernee: Technically I don't think the extension matters, but C header files typically have the `.h` extension.

doc/GUIDE.md
JornVernee: You mean add a comma like `e.g., ...`?

doc/GUIDE.md
mcimadamore, line 4: maybe add a link for FFM to the final JEP, so that readers who want to know more can do so?

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+archive of native functions and global variables. The user then has to look up the functions
+they want to call using a `SymbolLookup`, and finally 'link' the functions by using the
+`Linker::downcallHandle` method. Additionally, a client may need to create function pointer for
+Java functions using `Linker::upcallStub`, access global variables through the addresses
+returned by a lookup, and construct `MemoryLayout` instances for the structs they want to access.
+Jextract aims to automate many of these steps, so that a client can instead immediately start
+using a native library they are interested in.
```
mcimadamore, line 15: I suggest using `the native library` (because it's the one they are interested in). Also consider using plural (e.g. the native libraries).

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+Note that at this time, jextract only supports C header files. If you have a library written
+in another language, see the section on [other languages](#other-languages).
+
+## Running Jextract
+
+A native library typically has an `include` directory which contains all the header files
+that define the interface of the library, with one 'main' header file. Let's say we have a
+library called `mylib` stored at `/path/to/mylib` that has a directory `/path/to/mylib/include`
+with all the header files. And let's say that we have a shell open in the root directory
+of the Java project we're working on, which has an `src` source directory corresponding to
```
mcimadamore, line 29:
```suggestion
of the Java project we're working on, which has a `src` source directory corresponding to
```

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+
+This guide shows how to run the jextract tool, and how to use the Java code that it generates.
+The samples under [`samples`](samples) direcotry are also a good source of examples.
+
+Note that at this time, jextract only supports C header files. If you have a library written
+in another language, see the section on [other languages](#other-languages).
+
+## Running Jextract
+
+A native library typically has an `include` directory which contains all the header files
```
mcimadamore, line 25: Here I wonder if instead of using pseudo name for paths, we should use some more real-looking names, like `/usr/include` and `/usr/lib`, which are common in many NIX systems.

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+that define the interface of the library, with one 'main' header file. Let's say we have a
+library called `mylib` stored at `/path/to/mylib` that has a directory `/path/to/mylib/include`
+with all the header files. And let's say that we have a shell open in the root directory
```
mcimadamore, line 28: Maybe instead of `with all the header files` we could expand to `where the header files of that library are stored`

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+  called `mylib`. (The section on [library loading](#library-loading) discusses how is done)
+
+Besides these options, it is also possible to filter the output of jextract using one of the `--include-XXX` options
+that jextract has. See the section on [filtering](#filtering) for a more detailed overview. See also the full
+list of command line options [here](#command-line-option-reference).
+
+Jextract assumes that the version of a native library that a project uses is relatively stable.
+Therefore, jextract is intended to be run once, and then for the generated sources to be added to the project.
+Jextract only needs to be run again when the native library, or jextract itself are updated. (This is also
+the workflow that jextract itself uses for talking to the libclang library).
```
mcimadamore, line 63:
```suggestion
the workflow that jextract itself follows: jextract depends on the libclang (link?) native library in order to parse C sources).
```

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+
+### Library Loading
+
+When using the `--library <libspec>` option, the generated code internally uses [`SymbolLookup::libraryLookup`](https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/lang/foreign/SymbolLookup.html#libraryLookup(java.nio.file.Path,java.lang.foreign.Arena))
+to load libraries specified by `<libspec>`, after potentially mapping the name of the library to a platform dependent name using [`System::mapLibraryName`](https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/lang/System.html#mapLibraryName(java.lang.String)).
```
mcimadamore, line 68: potentially? Are there cases where mapping a library name is a no-op?

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+This means for instance that on Linux, when specifying `--library mylib` the bindings will try to load `libmylib.so` using the OS-specific
+library loading mechanism on Linux, which is [`dlopen`](https://man7.org/linux/man-pages/man3/dlopen.3.html).
+This way of loading libraries also relies on OS-specific search mechanisms to find the library file.
+On Linux the search path can be ammended using the `LD_LIBRARY_PATH` environment variable (see the documentation of `dlopen`).
```
mcimadamore, line 72:
```suggestion
On Linux the search path can be amended using the `LD_LIBRARY_PATH` environment variable (see the documentation of `dlopen`).
```

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+access functions, global variables, macros, enums, primitive typedefs, and layouts for
+builtin C types.
+
+The latter import statement imports all the other classes the jextract generates, which
+includes: classes representing structs or unions, function types, and struct or union
+typedefs.
+
+### Builtin Type Layouts
+
+For every jextract run, regardless of the contents of the library header files, jextract
```
mcimadamore, line 135: I'm not sure it's clear where these layouts definitions are introduced. I think the text should say that these are appended to the main header class.

mcimadamore: Perhaps you can add a comment:
```
// mylib_h.java
```
as you do in later examples?

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+public static int MY_MACRO() { ... }
+
+public static int A() { ... }
+public static int B() { ... }
+public static int C() { ... }
+```
+
+Note that the enum constants are exposed as top-level methods, rather than being nested
+inside a class called `MY_ENUM`, or through the use of a Java `enum`. This translation
+strategy mimics C's behavior of enum constants being accessible as a top-level declaration
```
mcimadamore, line 253: There's another reason as well for not using enums - which is that, in C, you can use bitwise operations to denote enum sets - e.g. `RED | GREEN` (and other more crazy stuff), which would not be possible using Java enums. At its core, a C enum is just a glorified int.

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+as well.
+
+Not all types of macros are supported though. Only macros that have a primitive numerical
+value, a string, or a pointer type are supported. Most notably, function-like macros are
```
mcimadamore, line 257: Several people wonder what to do when dealing with a library that is heavy in its use of function-like macros. There's two strategies here - the first is to replicate the code in the function-like macro in Java, using FFM. The second is to create a small C library which turns the function-like macro into a proper function, so that it can be used from FFM. Not sure we want to add this detail here (or maybe in an appendix).

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+            SegmentAllocator allocator) { ... } // 4
+
+    public static MemorySegment reinterpret(MemorySegment addr, Arena arena,
+            Consumer<MemorySegment> cleanup) { ... } // 6
+    public static MemorySegment reinterpret(MemorySegment addr, long elementCount,
+            Arena arena, Consumer<MemorySegment> cleanup) { ... } // 6
+}
+```
+
+There's a getter and setter for each field of the struct (1), which takes a pointer to a struct
```
mcimadamore, line 309: I see you didn't use bullet lists much, but I wonder if one would be helpful here instead of a single long para?

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+
+The `point` segment returned by `reinterpret` has exactly the size of one `Point` (for
+arrays of struct, use the `reinterpret` overload that takes an element count as well). The
+lifetime we associate with the segment is the lifetime denoted by `arena`, and when the
+arena is closed, we want to call `delete_point`, which we can do by passing a method
+reference to `delete_point` as a cleanup action when calling `reinterpret`.
+
+### Function Pointers
+
+If jextract finds a function pointer type in the header files it parses, it will generate
```
mcimadamore, line 392: More directly: `Jextract will generate a separate class for each function pointer type found in the header files it parses.`

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+
+try (Arena arena = Arena.ofConfined()) {
+    MemorySegment cb = callback_t.allocate((a, b) -> a * b, arena);
+    int result = call_me_back(cb);
+    System.out.println(result); // prints: 2
+} // 'cb' freed here
+```
+
+Here we use the lambda `(a, b) -> a * b` as the implementation of the `callback_t` instance
+we create using the call to `allocate`. This method returns an upcall stub like the ones
```
mcimadamore, line 448:
```suggestion
we create using `allocate`. This method returns an upcall stub like the ones
```

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+returned by the `java.lang.foreign.Linker::upcallStub` method. The `arena` argument denotes
```
mcimadamore, line 449: Should there be a link here - like we did for symbol lookup?

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+the lifetime of the upcall stub, meaning that the upcall stub will be freed when the arena
+is closed (after which the callback instance should no longer be called).
```
mcimadamore, line 451: "should not be called" seems to imply that if we call it something bad happens, but we can call it (e.g. it's up to our "goodwill"). "can no longer be called" seems more appropriate?

mcimadamore: Of course, if the native code stashes the function pointer in a variable somewhere, it would be possible for it to call the func pointer again. But what I meant is that, in principle, FFM should give you an exception when trying to call `call_me_back` again after the arena has been closed, or calling `callback_t::invoke` on a closed segment.

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+MemorySegment cb = get_callback();
+int result = callback_t.invoke(cb, 1, 2);
+System.out.println(result); // prints: 2
+```
+
+Here the `callback_t` instance we want to invoke is passed as the first argument to
+`invoke`, and then the `1` and `2` represent the arguments passed when calling the
+`callback_t` instance.
+
+Jextract generates function pointer classes like the `callback_t` class for function
```
mcimadamore, line 486: I know what you mean here - but I wonder if readers will get it w/o some kind of example.

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+pointers found in function parameter or return types, typedefs, or the types of variables
+(such as struct fields or global variables).
+
+### Variadic Functions
+
+Jextract handles variadic functions differently from regular functions. Variadic functions
+in C more or less like a template, where the calling convention changes based on the number
```
mcimadamore, line 493:
```suggestion
in C behave more or less like a template, where the calling convention changes based on the number
```

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+    System.out.println(Foo.bar.baz(bar));  // prints: 42
+}
+```
+
+In the above snippet, note that the load of the `baz` field value on the last line will
+'see' the update to the `bar` field of the `foo` instance on the line before.
+
+## Filtering
+
+Some libraries are incredibly large (such as a platform SDK), and we might not be
```
mcimadamore, line 660: Should we just say `windows.h` instead of the more vague "platform SDK" ?

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+interested in letting jextract generate code for the entire library. In cases like that,
+we can use jextract's `--include-XXX` command line options to only generate classes for
+the elements we specify.
+
+To allow for symbol filtering, `jextract` can generate a _dump_ of all the symbols
```
mcimadamore, line 665: I noticed that in these sections, refrerences to `jextract` use a code block, whereas in the rest of the document they don't. I don't have a strong preference for one or the other, but we should make that consistent.

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+that can be used by a client to access the native library.
+
+Interacting with native C code through the FFM API works
+by loading a native library (e.g. a `.so`/`.dll`/`.dylib` file), which is essentially an
+archive of native functions and global variables. The user then has to look up the functions
+they want to call using a `SymbolLookup`, and finally 'link' the functions by using the
+`Linker::downcallHandle` method. Additionally, a client may need to create function pointer for
+Java functions using `Linker::upcallStub`, access global variables through the addresses
+returned by a lookup, and construct `MemoryLayout` instances for the structs they want to access.
+Jextract aims to automate many of these steps, so that a client can instead immediately start
```
mcimadamore, line 14: If you want to avoid capitalization of `jextract` you can say `The jextract tool...`

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+```
+
+The value of these macros also affects the behavior of jextract. Therefore, jextract
+supports setting macro values on the command line using the `-D` or
+`--define-macro <macro>=<value>` option. For instance, we can use `-D MY_MACRO` to set
+the value of `MY_MACRO` in the above snippet to `1`, and trigger the first 'branch' of the
+compiler switch, thereby defining `int x = 42`.
+
+Please note that other header files included by jextract may also define macro values using
+the `#define` pre-processor directive. Therefore it is often important in which order
```
mcimadamore, line 781: There seems to be issues with this sentence. Possible suggestion:

`It is therefore important to notice the order in which header files are processed by a compiler, as feeding header files to jextract in the wrong order may result in weird errors due to missing macro definitions`

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+| `--use-system-load-library`                                  | libraries specified using `-l` are loaded in the loader symbol lookup (using either `System::loadLibrary`, or `System::load`). Useful if the libraries must be loaded from one of the paths in `java.library.path`.| 
+| `--output <path>`                                            | specify where to place generated files                       |
+| `--dump-includes <String>`                                   | dump included symbols into specified file (see below)        |
+| `--include-[function,constant,struct,union,typedef,var]<String>` | Include a symbol of the given name and kind in the generated bindings. When one of these options is specified, any symbol that is not matched by any specified filters is omitted from the generated bindings. |
+| `--version`                                                  | print version information and exit      
+
+### Additional clang options
+
+Users can also specify additional clang compiler options, by creating a file named
```
mcimadamore, line 808: Maybe here we should start by saying that `jextract` uses clang to parse C files. And if additional options need to be passed to the clang instance used by jextract...

doc/GUIDE.md
```diff
@@ -0,0 +1,822 @@
...
+`compile_flags.txt` in the current folder, as described
+[here](https://clang.llvm.org/docs/JSONCompilationDatabase.html#alternatives).
+
+## Other Languages
+
+As noted in the introduction, jextract currently only supports C header files, but many
```
mcimadamore, line 814: Should we say something about many libraries (e.g. libclang) providing a C interop wrapper around their C++ impl?

JornVernee: That's in the table below: `many C++ libraries have a C interface`

mcimadamore: whoops - true. I thought that row talked about manually writing a C wrapper around a C++ API.

doc/GUIDE.md
JornVernee: I think something like that is good to add here. This guide is meant to be comprehensive. If we think some information is useful in the use of jextract, let's add it to the guide.

doc/GUIDE.md
JornVernee: The `<libspec>` might be a path (using `:`), in which case we don't  do any mapping, right? I don't mind removing the `potentially` here though.

doc/GUIDE.md
JornVernee: It depends on how you read `src`. If you read S R C, then `an` is appropriate.

doc/GUIDE.md
JornVernee: I kind of wanted to hint at the fact that all library packages should have an `include` directory. The `/usr/include` directory can seem like a special NIX system path where libraries are installed, but that's more a quirk of the package manager which chooses to put the headers of all libraries in the same directory.

doc/GUIDE.md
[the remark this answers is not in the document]
JornVernee: Let's use italic, as that's the style we often use in the javadoc as well

doc/GUIDE.md
JornVernee: I think it would be appropriate here

doc/GUIDE.md
mcimadamore: In that case, I think it would be clearer if we broke up the sentence. E.g. 

```
When using the --library <libspec> option, the generated code internally uses SymbolLookup::libraryLookup
to load libraries specified by `<libspec>`. If `<libspec>` denotes a library name, the name is then mapped to a platform dependent name using System::mapLibraryName.
```

doc/GUIDE.md
JornVernee: Yes, I had `can no longer be called` before, but switched to `should not be called` since it's technically more correct. I think `can not longer be called` makes more sense though, in hindsight. Thanks

doc/GUIDE.md
JornVernee: I'll add some examples as well.

doc/GUIDE.md
JornVernee: I'll just switch to the plain jextract, as that's a little easier on the eyes. (This section was copied from the readme)

doc/GUIDE.md
JornVernee: I've added a leading sentence here in the latest version: `Jextract uses an embedded clang compiler (through libclang) to parse header files.`

doc/GUIDE.md
```diff
@@ -389,9 +404,8 @@ reference to `delete_point` as a cleanup action when calling `reinterpret`.
 
 ### Function Pointers
 
-If jextract finds a function pointer type in the header files it parses, it will generate
-a separate class for each of these. For instance, for a function pointer `typedef` like
-this:
+Jextract will generate a separate class for each function pointer type found in the header
```
mcimadamore, line 407:
```suggestion
Jextract generates a separate class for each function pointer type found in the header
```

doc/GUIDE.md
```diff
@@ -485,19 +500,29 @@ Here the `callback_t` instance we want to invoke is passed as the first argument
 
 Jextract generates function pointer classes like the `callback_t` class for function
 pointers found in function parameter or return types, typedefs, or the types of variables
-(such as struct fields or global variables).
+(such as struct fields or global variables):
+
+```c
```
mcimadamore, line 505: thanks!

mcimadamore (approved):

mcimadamore:
Great job! The guide reads great, and was long overdue :-)

nlisker:
I'm still reviewing, please allow another day before merging.

doc/GUIDE.md
```diff
@@ -0,0 +1,848 @@
...
+An invoker represents a particular _instantiation_ of a variadic function for a particular
+set of variadic parameter types. When the header files contain a variadic function like this:
+
+```c
+// mylib.h
+
+void foo_variadic(int x, ...);
+```
+
+Jextract doesn't generates a regular method, but a _class_, which represents the invoker:
```
nlisker, line 534: generates -> generate

doc/GUIDE.md
```diff
@@ -0,0 +1,848 @@
...
+Here we instantiate `foo_variadic` for 3 parameter types of the C type `int`. These parameter
+types essentially replace the `...` ellipsis in the C function type.
+
+We can call the instantiated invoker either by calling `apply`, which will box the arguments
+into an `Object[]`, or through the method handle returned by `handle()` which avoids the
+overhead of boxing.
+
+### Typedefs
+
+As mentioned before: typedefs are either translated as a `static final` memory layout fields
```
nlisker, line 576: before: -> before,

doc/GUIDE.md
```diff
@@ -0,0 +1,848 @@
...
+`MyInt` is a `typedef` of the primitive type `int`, so it is translated by jextract as a
+`static final` layout field in the main header class:
+
+```java
+// mylib_h.java
+
+public static final OfInt MyInt = mylib_h.C_INT;
+```
+
+The `MyPoint` `typedef` on the other hand is a typedef for a struct, so it is translated
```
nlisker, line 603: "`typedef` on the other hand" -> "`typedef`, on the other hand,"

doc/GUIDE.md
```diff
@@ -0,0 +1,848 @@
...
+public class MyPoint extends Point { }
+```
+
+Through static inheritance, all the methods in the `Point` class are available through the
+`MyPoint` class as well.
+
+### Nested Types
+
+C allows variable declarations to have an inline anonymous type. Jextract handles in
+particular cases where a struct's field has an inline type specially. For instance, if we
```
nlisker, line 618: "Jextract handles in particular cases where a struct's field has an inline type specially."

I don't understand this sentence. Does Jextract handle nested types only when a struct's field has an inline type?

doc/GUIDE.md
```diff
@@ -0,0 +1,848 @@
...
+    struct {
+        int baz;
+    } bar; // field of Foo
+
+    void (*cb)(void);
+};
+```
+
+Jextract generates a _nested_ struct and function pointer class for the `bar` and `cb`
+fields, _inside of_ the class it generates for the `Foo` struct itself:
```
nlisker, line 634: fields, -> fields

doc/GUIDE.md
```diff
@@ -0,0 +1,848 @@
...
+
+--include-struct Foo                  # header: /workspace/myproj/mylib.h
+--include-struct Point                # header: /workspace/myproj/mylib.h
+--include-typedef callback_t          # header: /workspace/myproj/mylib.h
+--include-function call_me_back       # header: /workspace/myproj/mylib.h
+...
+```
+
+The include options in this file can then be edited down to a set of symbols that is
+desired, for instance using other command line tools such as `grep` or `Select-String`,
```
nlisker, line 713: instance -> instance,

doc/GUIDE.md
```diff
@@ -0,0 +1,848 @@
...
+```c
+#ifdef MY_MACRO
+int x = 42;
+#else
+int x = 0;
+#endif
+```
+
+The value of these macros also affects the behavior of jextract. Therefore, jextract
+supports setting macro values on the command line using the `-D` or
```
nlisker, line 800: Perhaps mentioning the distinctions between these macros and the ones described in the Constants section more specifically will be helpful?

nlisker (commented):
An excellent guide. I read through all of it as someone who already used jextract and left some minor grammar comments that I found to help with the flow, but I would like to point out something more confusing with the structure.

The first part is about running jextract, including a short review of an example, followed by an explanation about loading libraries. It looks like this section explains the arguments/options passed to the tool, but it skips the discussion about the macros option (appears much later in Preprocessor Definitions) and other options.

The second part is about the code that is generated, but then there is an explanation about `--header-class-name` and `--target-package`, which should be part of the previous part. I would expect that by the time the user gets to learning about the generated Java code, they would know what commands generated it.

Also:
* Did you leave out "Bitfields" on purpose?
* Can a ToC be generated?

doc/GUIDE.md
```diff
@@ -0,0 +1,848 @@
...
+libraries also relies on OS-specific search mechanisms to find the library file. On Linux
+the search path can be amended using the `LD_LIBRARY_PATH` environment variable (see the
+documentation of `dlopen`). On Mac the relevant environment variable is `DYLD_LIBRARY_PATH`,
+and on Windows the variable is `PATH`. Though, for the latter the overall library search
+mechanism is entirely different (described [here](https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order)).
+When using the HotSpot JVM, the `-Xlog:library` option can als be use to log where the JVM
+is trying to load a library from, which can be useful to debug a failure to load a library.
+
+The `<libspec>` argument of the `--library` option can either be a library name, such as,
+`mylib` which will, be mapped to a platform specific name using [`System::mapLibraryName`](https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/lang/System.html#mapLibraryName(java.lang.String)), or a path to a library file (either relative or
```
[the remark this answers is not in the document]
mcimadamore: I'd also drop the comma after `such as`

doc/GUIDE.md
```diff
@@ -0,0 +1,848 @@
...
+
+Where `org.mypackage` is the package into which jextract put the generates source files
+(using `--target-package`/`-t`).
+
+The former import statement will import all the static functions and fields from the class
+that jextract generates for the main header file of the library. This includes methods to
+access functions, global variables, macros, enums, primitive typedefs, and layouts for
+builtin C types.
+
+The latter import statement imports all the other classes the jextract generates, which
```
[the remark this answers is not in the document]
mcimadamore: or, `generated by jextract`

doc/GUIDE.md
```diff
@@ -0,0 +1,848 @@
...
+try (Arena arena = Arena.ofConfined()) {
+    MemorySegment point = Point.allocate(arena);
+    Point.x(point, 10);
+    Point.y(point, 5);
+    // ...
+}
+```
+
+For working with arrays of structs, we can use the `allocateArray` method which accepts an
+additional element count, indicating the length of the array:
```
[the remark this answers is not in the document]
mcimadamore: I find the way it's currently written fine (and better than the suggested one)

mcimadamore:
> Did you leave out "Bitfields" on purpose?

Jextract does not support bitfields. That's why there's no section for them :-)

doc/GUIDE.md
```diff
@@ -0,0 +1,848 @@
...
+Note that at this time, jextract only supports C header files. If you have a library written
+in another language, see the section on [other languages](#other-languages).
+
+## Running Jextract
+
+A native library typically has an `include` directory which contains all the header files
+that define the interface of the library, with one _main_ header file. Let's say we have a
+library called `mylib` stored at `/path/to/mylib` that has a directory `/path/to/mylib/include`
+where the header files of that library are stored. And let's say that we have a shell open
+in the root directory of the Java project we're working on, which has an `src` source
```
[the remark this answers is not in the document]
mcimadamore: I believe this was discussed here:
https://github.com/openjdk/jextract/pull/231#discussion_r1559141069

nlisker:
> > Did you leave out "Bitfields" on purpose?
> 
> Jextract does not support bitfields (due to deficiencies in the libclang API). That's why there's no section for them :-)

Then maybe mention that :)

doc/GUIDE.md
```diff
@@ -0,0 +1,848 @@
...
+
+## Tracing
+
+It is sometimes useful to inspect the parameters passed to a native call, especially when
+diagnosing application bugs and/or crashes. The code generated by the jextract tool
+supports _tracing_ of native calls, that is, parameters passed to native calls can be
+printed on the standard output.
+
+To enable the tracing support, just pass the `-Djextract.trace.downcalls=true` flag to the
+launcher used to start the application. Below we show an excerpt of the output when
```
nlisker, line 757: I would mention that this is passed as a VM argument, not as a program argument.

JornVernee:
> The first part is about running jextract, including a short review of an example, followed by an explanation about loading libraries. It looks like this section explains the arguments/options passed to the tool, but it skips the discussion about the macros option (appears much later in Preprocessor Definitions) and other options.
> 
> The second part is about the code that is generated, but then there is an explanation about `--header-class-name` and `--target-package`, which should be part of the previous part. I would expect that by the time the user gets to learning about the generated Java code, they would know what commands generated it.

Yes. The intent of this was to not bog down the intro with information that is not used that often, so people can get started quickly. But, in hindsight, I think it is better to just put all the discussion about command line arguments in 1 place. (Especially some of the info in the section about `-D` is important to get jextract to 'work'). I'll re-arrange the sections a bit.

doc/GUIDE.md
```diff
@@ -0,0 +1,848 @@
...
+from the parsed header files.
+
+### Global Variables
+
+For a global variable declaration in a header file like this:
+
+```c
+// mylib.h
+
+int bar;
```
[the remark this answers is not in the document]
JornVernee: For jextract it doesn't matter

doc/GUIDE.md
```diff
@@ -0,0 +1,848 @@
...
+```
+
+We can call this function from Java as follows:
+
+```java
+// Main.java
+
+try (Arena arena = Arena.ofConfined()) {
+    MemorySegment cb = callback_t.allocate((a, b) -> a * b, arena);
+    int result = call_me_back(cb);
```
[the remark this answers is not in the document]
JornVernee: It's defined in the C file in the snippet above.

doc/GUIDE.md
JornVernee: C doesn't allow other kinds of nested types, so yeah

JornVernee: I'll just remove the 'Jextract handles in particular cases where a struct's field has an inline type specially.' part here

doc/GUIDE.md
JornVernee: Hmm, not sure what to say. There are no real distinctions, but jextract only generates an accessor if a macro is defined in the header files, not if it is only set on the command line with `-D`. (I'll add that at least)

doc/GUIDE.md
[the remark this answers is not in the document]
nlisker: But we can't call the c function directly. Are there more java bindings that jextract generates?

JornVernee:
> Can a ToC be generated?

It should be generated automatically by GitHub

JornVernee:
> > > Did you leave out "Bitfields" on purpose?
> > 
> > 
> > Jextract does not support bitfields (due to deficiencies in the libclang API). That's why there's no section for them :-)
> 
> Then maybe mention that :)

I've added a section about unsupported features. (also to explain some of the warnings a user may see)

doc/GUIDE.md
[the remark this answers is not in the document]
JornVernee: Sorry, the example assumes that that the `call_me_back` function is also defined in the header file that jextract extracted, so a Java method would be generated for it.

doc/GUIDE.md
```diff
@@ -0,0 +1,848 @@
...
+public static MethodHandle foo$handle() { ... } // 4
+```
+
+First and foremost, there is a static wrapper method that is generated that can be used to
+call the C function (1). Besides that, there are also several accessors that return
+additional meta-data for the method: the function's address (2), the function descriptor
+(3), and the method handle returned by the FFM linker (4), which is used to implement the
+static wrapper method (1).
+
+The parameter types and return type of this method depend on the carrier types of the
```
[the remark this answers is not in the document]
JornVernee: 'carrier type` is indeed an FFM concept. I've added a link to `FunctionDescriptor::toMethodType` which explains how these are derived.

doc/GUIDE.md
```diff
@@ -0,0 +1,848 @@
...
+// mylib.h
+
+void foo(int x);
+```
+
+Jextract will generate the following set of methods for this function, in the main header
+class it generates:
+
+```java
+// mylib_h.java
```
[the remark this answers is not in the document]
JornVernee: Don't think it's that important to mention. The comment is mostly there for consistency, as the text above already says that this is the class that jextract generates.

doc/GUIDE.md
```diff
@@ -103,17 +144,6 @@ In the following section we'll go over examples of declarations that might be fo
...
-jextract. For example, if the header file is named `mylib.h`, then the derived class name
-will be `mylib_h`. The class name can be set explicitly using the `--header-class-name`
-option as well.
-
-Besides the main header class that is generated, jextract also generates several
-separate files for certain declarations in the C header file. Namely, structs, unions,
-function pointer types, and typedefs of struct or union types result in additional files
-being generated. We will look at those in more detail in this section.
-
 Most of the methods that jextract generates are `static`, and are designed to be imported
```
mcimadamore, line 147: Uhm - I think the paras you moved fitted nicely in this introductory section

doc/GUIDE.md
```diff
@@ -193,12 +223,15 @@ public static MethodHandle foo$handle() { ... } // 4
 
 First and foremost, there is a static wrapper method that is generated that can be used to
 call the C function (1). Besides that, there are also several accessors that return
-additional meta-data for the method: the function's address (2), the function descriptor
-(3), and the method handle returned by the FFM linker (4), which is used to implement the
-static wrapper method (1).
-
-The parameter types and return type of this method depend on the carrier types of the
-layouts that make up the function descriptor of the function, which is itself derived
+additional meta-data for the method: the function's address, represented as a
```
mcimadamore, line 226: Should we use a numbered list here too?

doc/GUIDE.md
```diff
@@ -266,9 +299,13 @@ Not all types of macros are supported though. Only macros that have a primitive
 value, a string, or a pointer type are supported. Most notably, function-like macros are
 not supported by jextract. For function-like macros, alternatives include re-writing the
 code inside the macro in Java, using the FFM API, or writing a small C library which wraps
-the function-like macro in a proper exported C function, that can then be linked against
+the function-like macro in a proper exported C function that can then be linked against
 through the FFM API.
 
+Note that for macros, jextract only generates an accessor when it sees a macro definition,
+like the one in the example, in the header files it parses. When a macro is only defined
+using `-D` on the command line, not accessor will be generated.
```
mcimadamore, line 307:
```suggestion
using `-D` on the command line, no accessor will be generated.
```

doc/GUIDE.md
```diff
@@ -266,9 +299,13 @@ Not all types of macros are supported though. Only macros that have a primitive
 value, a string, or a pointer type are supported. Most notably, function-like macros are
 not supported by jextract. For function-like macros, alternatives include re-writing the
 code inside the macro in Java, using the FFM API, or writing a small C library which wraps
-the function-like macro in a proper exported C function, that can then be linked against
+the function-like macro in a proper exported C function that can then be linked against
 through the FFM API.
 
+Note that for macros, jextract only generates an accessor when it sees a macro definition,
+like the one in the example, in the header files it parses. When a macro is only defined
```
mcimadamore, line 306:
```suggestion
like the one in the example, in the header files it parses. When a macro is defined
```

doc/GUIDE.md
JornVernee: I think they could go in either section? But, this way, we can discuss the `--header-class-name` option in the section about running, which I think makes more sense.

doc/GUIDE.md
JornVernee: I don't think it's possible to start a markdown list at `2`, but I will try

doc/GUIDE.md
```diff
@@ -835,6 +843,30 @@ can also specify additional clang compiler options, by creating a file named
 `compile_flags.txt` in the current folder, as described
 [here](https://clang.llvm.org/docs/JSONCompilationDatabase.html#alternatives).
 
+## Unsupported Features
+
+There are several elements for which jextract can not generate bindings:
```
mcimadamore, line 849: element seems a bit vague, maybe "C program elements" could be better, although in this cases I suppose most developers will think about the term "features". Consider just having a bullet list and skip the first statement (after all, the title says it all). Note also the list is provided here:
https://jdk.java.net/jextract/
(maybe we should drop it from there in the future, to avoid duplication?)

doc/GUIDE.md
```diff
@@ -835,6 +843,30 @@ can also specify additional clang compiler options, by creating a file named
...
+  struct Foo;
+  ```
+
+  Jextract is not able to generate the regular `Foo` class. You will see a warning about
+  these structs being skipped, such as:
+
+  ```txt
+  WARNING: Skipping Foo (type Declared(Foo) is not supported)
+  ```
+
```
mcimadamore, line 870: Bitfields missing?

doc/GUIDE.md
JornVernee: `only` is relevant here. If a macro is defined using `#define` _and_ on the command line, jextract will generate an accessor.

doc/GUIDE.md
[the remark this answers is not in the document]
nlisker: I would mentioned that. After all, this tool is meant to save Java developers from knowing the intricacies of the native language.

doc/GUIDE.md
JornVernee: Bitfields is there. I'll add an item for types > 64 bits

doc/GUIDE.md
```diff
@@ -58,31 +58,72 @@ Besides these options, it is also possible to filter the output of jextract usin
...
 Jextract assumes that the version of a native library that a project uses is relatively stable.
 Therefore, jextract is intended to be run once, and then for the generated sources to be added to the project.
-Jextract only needs to be run again when the native library, or jextract itself are updated. (This is also
-the workflow that jextract itself follows: jextract depends on the [libclang](https://clang.llvm.org/docs/LibClang.html)
-native library in order to parse C sources).
+Jextract only needs to be run again when the native library, or jextract itself are updated.
+(This is also the workflow that jextract itself follows: jextract depends on the
+[libclang](https://clang.llvm.org/docs/LibClang.html) native library in order to parse C sources).
+
+### Preprocessor Definitions
```
mcimadamore, line 79: IMHO both this and the following section is not important enough to appear so early in the document. If we went straight to "Using The Code Generated By Jextract" we would not lose anything - then after that section, we can provide more details about various aspects/options.

doc/GUIDE.md
mcimadamore: I see bitfields now, but we should say something for e.g. `long double` on Linux

doc/GUIDE.md
JornVernee: Not sure... I know what you mean, but knowing how to run the generated code seems fairly important. It seems that that is something someone might want to try out after the first example. Unfortunately, this requires understanding library loading, which is fairly complex.

Also, the comment about feeding jextract the right header file seems pretty important since user may run into errors when running jextract otherwise (we've seen many questions about that, e.g. when someone tries to extract some arbitrary Windows header instead of `Windows.h`).

But also, I realized that this guide may not just be used by first-time readers, but also as a reference by more experienced users, in which case it seems nice to have info about command line flags mostly in one place.

doc/GUIDE.md
mcimadamore: On a second look, it makes sense in this updated version.

doc/GUIDE.md
JornVernee: Looks like this doesn't work unfortunately.

JornVernee: No, wait, got it working after all.

doc/GUIDE.md
mcimadamore: While it's true that knowing which header to extract is important, I think having a whole section on "preprocessor definitions" feels like having a very cornery section in a place where in reality you want something more general. E.g. extracting windows.h is tricky, because of order dependencies, but I'm not sure the user needs to be aware of -D here. Just mentioning that headers are complex beasts and can depend on each other in unpredictable ways (and refer to the doc of the library that needs to be extracted) should be good enough guidance at this point.

As for library loading I'm not sure... again, yes, we probably need to say something like "you need to ensure the library you want to call is loaded first" - but I'm not sure we need the full shebang of library loading options. In other places you added references to stuff that is further expanded upon in later sections - e.g. 

```
Besides these options, it is also possible to filter the output of jextract using one of the --include-XXX options that jextract has. See the section on [filtering](https://github.com/JornVernee/jextract/blob/Guide/doc/GUIDE.md#filtering) for a more detailed overview.
```

Seems like something similar is required here.

doc/GUIDE.md
```diff
@@ -231,6 +231,7 @@ public static MethodHandle foo$handle() { ... } // 4
 First and foremost, there is a static wrapper method that is generated that can be used to
 call the C function (1). Besides that, there are also several accessors that return
 additional meta-data for the method:
+
```
mcimadamore, line 234: You could also add an item for (1) instead of having separate text above?

JornVernee: Done

mcimadamore: Now veering into nitpicks: but I note that we have an asymmetry between structs and functions. In functions we try to call out the function accessor (probably on the basis that this is the symbol the user is likely to be looking for). But for structs we just list all the members one after the other (even though the most likely is probably (1) - e.g. getter/setter). I think the document could be improved (optionally) by rectifying this one way or the other.

doc/GUIDE.md
JornVernee: Still not sure. The section on pre-processor definitions is already fairly small, and, in some rare cases, `-D` may be required for extraction to work as well. Do we really want to put ~half this section elsewhere? It seems this information is relevant to 'Running Jextract' as well.

For library loading, a user needs to know what happens to the string they pass to `--library`, which requires understanding the distinction between library names/paths, and the fact that the user needs to set `LD_LIBRARY_PATH`/`DYLD_LIBRARY_PATH`/`PATH`. Even for someone who's experienced with native library loading, that seems like important information?

I think in both cases, we would end up splitting just part of the information into a separate section. Even if not ideal, I think it's better to keep all the info together. So, if we can't get away with moving these sections out-of-line completely, which I don't think we can, I'm inclined to leave things as they are in this area.

Maybe we can improve the situation by have a very basic hello world example up front?

mcimadamore: IMHO is not about moving "half sections", but about writing a section that is correct for the purpose in which it appears in the document. We seem to have two requirements for users getting started with jextract:
* explain that the header on which it is run matters (this can literally be one sentence)
* explain that libraries must be loaded (I think here we can get away by providing something that works in the 80% use cases, plus some comment on LD_LIBRARY_PATH, and refer everything else to a more specialized section. E.g. I don't think here we should be talking about different lookups, in which orders symbols are looked up etc.).

This way you can have a document that serves both "green" developers and more experienced one (the latter will want to keep going after the section illustrating how to use bindings). At least, IMHO.

mcimadamore: I personally like the new changes a lot! I find the "running jextract" section very informative, providing all the major info w/o overstaying its welcome. Also it points to several other sections in the doc, so it acts a bit as a mental map, which is good.

I have some reservation about "unsupported features" being down in "Advanced" - should that be the last section in "Using jextract bindings" instead? (e.g. first we show what works, and then if the reader was expecting something else, we have a section for that too). Or, maybe pull it more "up front". Right now it seems kind of buried in other stuff, even thought it seems generally useful.

JornVernee: Ok, I'll keep it like this then.

I'll move the unsupported feature to the end of the 'using' section.

JornVernee:
I think I've addressed all review comments. Please have another look

doc/GUIDE.md
JornVernee: Ok, removed.

nlisker (commented):
Looks very good. Left some minor comments. Can be integrated after they are resolved.

doc/GUIDE.md
```diff
@@ -0,0 +1,910 @@
...
+    public static MemorySegment allocate(SegmentAllocator allocator) { ... } // 4
+    public static MemorySegment allocateArray(long elementCount,
+            SegmentAllocator allocator) { ... } // 4
+
+    public static MemorySegment reinterpret(MemorySegment addr, Arena arena,
+            Consumer<MemorySegment> cleanup) { ... } // 6
+    public static MemorySegment reinterpret(MemorySegment addr, long elementCount,
+            Arena arena, Consumer<MemorySegment> cleanup) { ... } // 6
+}
```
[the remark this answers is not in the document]
JornVernee: static methods don't need to be `final`. I think that may be an artifact of some code sharing we do internally.

doc/GUIDE.md
```diff
@@ -0,0 +1,910 @@
...
+not supported by jextract. For function-like macros, alternatives include re-writing the
+code inside the macro in Java, using the FFM API, or writing a small C library which wraps
+the function-like macro in a proper exported C function that can then be linked against
+through the FFM API.
+
+Note that for macros, jextract only generates an accessor when it sees a macro definition,
+like the one in the example, in the header files it parses. When a macro is defined
+using `-D` on the command line, no accessor will be generated.
+
+### Structs & Unions
```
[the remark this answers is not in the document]
JornVernee: It's exactly the same for both (I'll mention that as well)

doc/GUIDE.md
```diff
@@ -0,0 +1,910 @@
...
+documentation of the library that you're trying to use to find out which header file should
+be included/passed to jextract.
+
+### Library Loading
+
+When using the `--library <libspec>` option, the generated code internally uses [`SymbolLookup::libraryLookup`](https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/lang/foreign/SymbolLookup.html#libraryLookup(java.nio.file.Path,java.lang.foreign.Arena))
+to load libraries specified by `<libspec>`. If `<libspec>` denotes a library name, the
+name is then mapped to a platform dependent name using [`System::mapLibraryName`](https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/lang/System.html#mapLibraryName(java.lang.String)).
+This means, for instance, that on Linux, when specifying `--library mylib`, the bindings will
+try to load `libmylib.so` using the OS-specific library loading mechanism on Linux, which
```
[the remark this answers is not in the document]
JornVernee: No, `mapLibraryName` adds the `lib` prefix on Linux

doc/GUIDE.md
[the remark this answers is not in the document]
JornVernee: I put this:

```
We can call this function from Java as follows (assuming this function is also exported
through the header files passed to jextract)
```

doc/GUIDE.md
```diff
@@ -0,0 +1,910 @@
...
+- `--include-dir /path/to/mylib/include` specifies a header file search directory, which
+   is used to find header files included through `#include` in the main header file.
+- `--output src` specifies the root directory for the output. This matches the root package
+  of the project's source directory.
+- `--target-package org.jextract.mylib` specifies the target package to which the generated
+  classes and interfaces will belong. (Note that jextract will automatically create the
+  directories representing the package structure under the `src` directory specified
+  through `--output`)
+- `--library mylib` tells jextract that the generated bindings should load the library
+  called `mylib`. (The section on [library loading](#library-loading) discusses how is done)
```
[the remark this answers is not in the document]
JornVernee: This should be have "this is"

doc/GUIDE.md
```diff
@@ -0,0 +1,910 @@
+# Jextract Guide
+
+The jextract tool parses header (.h) files of native libraries, and generates Java code,
+called _bindings_, which use the [Foreign Function and Memory API](https://openjdk.org/jeps/442)
+(FFM API) under the hood, that can be used by a client to access the native library.
+
+Interacting with native C code through the FFM API works
+by loading a native library (e.g., a `.so`/`.dll`/`.dylib` file), which is essentially an
+archive of native functions and global variables. The user then has to look up the functions
+they want to call using a `SymbolLookup`, and finally _link_ the functions by using the
```
mcimadamore, line 10: There's some classes and methods referenced here - should we use links? (don't have strong opinion)

doc/GUIDE.md
```diff
@@ -0,0 +1,910 @@
...
+`Linker::downcallHandle` method. Additionally, a client may need to create function pointers for
+Java functions using `Linker::upcallStub`, access global variables through the addresses
+returned by a lookup, and construct `MemoryLayout` instances for the structs they want to access.
+The jextract tool aims to automate many of these steps, so that a client can instead
+immediately start using the native libraries they are interested in.
+
+This guide shows how to run the jextract tool, and how to use the Java code that it generates.
+The samples under the [`samples`](samples) direcotry are also a good source of examples.
+
+Note that at this time, jextract only supports C header files. If you have a library written
```
mcimadamore, line 20: should we say "jextract (and FFM) only support..." ?

doc/GUIDE.md
```diff
@@ -0,0 +1,910 @@
...
+header file which `#include`s these header files, and then this new header file can be
+passed to jextract.
+
+The library name specified to `--library` will be mapped to a platform specific library
+file name, and should be findable through the OS's library search mechanism, typically by
+specifying the library's containing directory on `LD_LIBRARY_PATH` (Linux),
+`DYLD_LIBRARY_PATH` (Mac), or `PATH` (Windows). See the section on
+[library loading](#library-loading) for more information.
+
+Besides these options, it is also possible to filter the output of jextract using one of the `--include-XXX` options
```
mcimadamore, line 70:
```suggestion
Besides these options, it is also possible to filter the output of jextract using one of the `--include-XYZ` options
```

doc/GUIDE.md
```diff
@@ -0,0 +1,910 @@
...
+declarations:
+
+```c
+// mylib.h
+
+#define MY_MACRO 42
+enum MY_ENUM { A, B, C };
+```
+
+Jextract will generate a set of simple getter methods to access the constant values of the
```
mcimadamore, line 236: Nit: maybe omit the word `set` - e.g. just "generate simple getter methods...". Seeing "set" so close to "getter" might be confusing.

doc/GUIDE.md
```diff
@@ -0,0 +1,910 @@
...
+
+In the above snippet, note that the load of the `baz` field value on the last line will
+_see_ the update to the `bar` field of the `foo` instance on the line before.
+
+### Unsupported Features
+
+Finally, there are some features that jextract does not support, listed below:
+
+- Certain C types bigger than 64 bits (e.g. `long double` on Linux).
+- Function-like macros (as mentioned in the [section on constants](#constants-macros--enums))
```
mcimadamore, line 682: I wonder if here we got the reference backwards - e.g. perhaps the section on constants should just say that function-like macros are not supported and point here, then here we could say what workarounds could be used in more details?

doc/GUIDE.md
```diff
@@ -0,0 +1,910 @@
...
+with Java `enum`s.
+
+Not all types of macros are supported though. Only macros that have a primitive numerical
+value, a string, or a pointer type are supported. Most notably, function-like macros are
+not supported by jextract. For function-like macros, alternatives include re-writing the
+code inside the macro in Java, using the FFM API, or writing a small C library which wraps
+the function-like macro in a proper exported C function that can then be linked against
+through the FFM API.
+
+Note that for macros, jextract only generates an accessor when it sees a macro definition,
```
mcimadamore, line 262: Feels like here we need a link to the preprocessor section

doc/GUIDE.md
```diff
@@ -0,0 +1,910 @@
...
+returned from native code. Let's say we have a C function that creates an instance
+of a `Point`, and returns a pointer to it, as well as a function that deletes a
+point, given a pointer:
+
+```c
+struct Point* new_point(void);
+void delete_point(struct Point* ptr);
+```
+
+The pointer that is returned by the corresponding method that jextract generates
```
mcimadamore, line 376: Consider streamlining this by saying "The pointer that is returned by new_point does not have..."

doc/GUIDE.md
```diff
@@ -0,0 +1,910 @@
...
+for this function does not have the correct bounds or lifetime associated with it.
+These things are not possible to figure out automatically (for instance, a pointer
```
mcimadamore, line 378:
```suggestion
These things cannot be inferred (for instance, a pointer
```

doc/GUIDE.md
```diff
@@ -0,0 +1,910 @@
...
+        int apply(int x, int y);
+    }
+
+    public static FunctionDescriptor descriptor() { ... }
+    public static MemorySegment allocate(callback_t.Function fi, Arena arena) { ... }
+    public static int invoke(MemorySegment funcPtr,int x, int y) { ... }
+}
+```
+
+We again have a meta-data accessor for the function descriptor (`descriptor()`). There's
```
mcimadamore, line 426: Maybe here using numbers in the code and a numbered list would be (a) more consistent and (b) more readable

doc/GUIDE.md
```diff
@@ -0,0 +1,910 @@
...
+    MemorySegment bar2 = Foo.bar.allocate(arena);
+    Foo.bar.baz(bar2, 42);
+    Foo.bar(foo, bar2);  // copies bar2 into foo
+    System.out.println(Foo.bar.baz(bar));  // prints: 42
+}
+```
+
+In the above snippet, note that the load of the `baz` field value on the last line will
+_see_ the update to the `bar` field of the `foo` instance on the line before.
+
```
mcimadamore, line 676: Perhaps we could also add a section (maybe in a followup PR) on fields/globals with array types, as jextract generates extra metadata (e.g. dimensions) and the Java type used for these accesses is MemorySegment (with bulk copy support on `set`)

doc/GUIDE.md
```diff
@@ -0,0 +1,910 @@
...
+  ```txt
+  WARNING: Skipping Foo (type Declared(Foo) is not supported)
+  ```
+
+## Advanced
+
+### Preprocessor Definitions
+
+C header files are processed by a pre-processor by a compiler before they are inspected
```
mcimadamore, line 706: I think you mean pre-processor, and remove "by a compiler" ?

doc/GUIDE.md
mcimadamore: Errno and endianness also come to mind in terms of unsupported stuff

mcimadamore (approved):
I've reviewed the document again. It looks very good, and the structure of the documents strikes me as just right (well done!). I've left picky comments (e.g. 99% review), which I consider optional. I'm ok with the PR being pushed in its current form.

doc/GUIDE.md
JornVernee: I think so. This is introductory, so someone may want to read up on the FFM API before diving deeper in  jextract

doc/GUIDE.md
JornVernee: Yes, good point. This was missed

JornVernee: let's save this for a followup, and get this stake in the ground first :) (it will probably also be easier to review a small section)

JornVernee:
I've address the last of the review comments. I've also move all the javadoc links out of line, since they are so long (and it might be nice to have them in one place, should we need to update them in the future).

I'll integrate this PR if there are no further comments.

mcimadamore (approved):

PR #241, opened 2024-04-30
7903720: Change layout of jextract image

JornVernee:
See JBS issue for problem description.

This patch changes the layout of the jextract image we produce. The new image has the 2 launcher scripts in the root of the image, with a nested `runtime` folder containing the jlink runtime image.

I've added both launcher scripts to the repo under the `src/main` directory, and they are copied to the image during the build. The script is somewhat simplified by removing a blank variable that was not being used. I also took the opportunity to address an issue with the windows launcher script turning off `echo`. Instead it now avoids output using the `@` prefix for each command in the script.

I've tested that the scripts work (using `jextract --help`) on both Linux and Windows.

---------
### Progress

### Issue
 * [CODETOOLS-7903720](https://bugs.openjdk.org/browse/CODETOOLS-7903720): Change layout of jextract image (**Enhancement** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/241/head:pull/241` \
`$ git checkout pull/241`

`$ git checkout pull/241` \
`$ git pull https://git.openjdk.org/jextract.git pull/241/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 241`

`$ git pr show -t 241`

</details>

<a href="https://git.openjdk.org/jextract/pull/241.diff">https://git.openjdk.org/jextract/pull/241.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/241#issuecomment-2085560502)


src/main/jextract
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+DIR=`dirname $0`
+$DIR/runtime/bin/java -m org.openjdk.jextract/org.openjdk.jextract.JextractTool "$@"
```
mcimadamore, line 3: What do you think about adding a dependency on some shell variable (e.g. JEXTRACT_JAVA_OPTIONS) to pass extra arguments to jextract's JVM? I recall reaching for something like this e.g. to enable debug mode? (But, totally optional, more asking for an opinion)

mcimadamore (approved):

src/main/jextract
JornVernee: Yes, this seems like a good idea.

Previously I would just modify the launch script.

JornVernee:
I've made a few more changes:

1. Implemented Maurizio's suggestion to allow passing options to the launcher through a shell variable
2. Set POSIX executable permissions for the copied script
3. Disable macos GHA runs. It looks like they switched their runner to arm64, so we can't use x64 llvm any more it seems.

mcimadamore (approved):

PR #244, opened 2024-05-20
7903727: Remove the reliance on String Templates feature

nizarbenalla:
This PR aims to replace the usage of string templates with `String::format`, since there will be no string template feature in JDK 23.
I tried to keep similar indentation and convert them in-place, to make reviewing the changes easier.

---------
### Progress

### Issue
 * [CODETOOLS-7903727](https://bugs.openjdk.org/browse/CODETOOLS-7903727): Remove the reliance on String Templates feature (**Bug** - P4)

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - **Reviewer**) ⚠️ Review applies to [042cc415](https://git.openjdk.org/jextract/pull/244/files/042cc4151ac47c0c540657d1f2157991bc2607a5)
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to [042cc415](https://git.openjdk.org/jextract/pull/244/files/042cc4151ac47c0c540657d1f2157991bc2607a5)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/244/head:pull/244` \
`$ git checkout pull/244`

`$ git checkout pull/244` \
`$ git pull https://git.openjdk.org/jextract.git pull/244/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 244`

`$ git pr show -t 244`

</details>

<a href="https://git.openjdk.org/jextract/pull/244.diff">https://git.openjdk.org/jextract/pull/244.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/244#issuecomment-2120047970)


src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
```diff
@@ -144,27 +142,27 @@ void appendIndentedLines(String s) {
     }
 
     final void emitDefaultConstructor() {
-        appendIndentedLines(STR."""
+        appendIndentedLines(String.format("""
```
mcimadamore, line 145: Suggestion - tweak the `appendIndentedLines` to call `String::format` inside that method, so that all clients don't have to call `String::format` manually from outside.

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
```diff
@@ -192,11 +190,11 @@ String layoutString(Type type, long align) {
         return switch (type) {
             case Primitive p -> primitiveLayoutString(p, align);
             case Declared d when Utils.isEnum(d) -> layoutString(((Constant)d.tree().members().get(0)).type(), align);
-            case Declared d when Utils.isStructOrUnion(d) -> alignIfNeeded(STR."\{JavaName.getFullNameOrThrow(d.tree())}.layout()", ClangAlignOf.getOrThrow(d.tree()) / 8, align);
-            case Delegated d when d.kind() == Delegated.Kind.POINTER -> alignIfNeeded(STR."\{runtimeHelperName()}.C_POINTER", 8, align);
+            case Declared d when Utils.isStructOrUnion(d) -> alignIfNeeded(String.format("%s.layout()", JavaName.getFullNameOrThrow(d.tree())), ClangAlignOf.getOrThrow(d.tree()) / 8, align);
```
mcimadamore, line 193: Same thing here - if all calls to `alignIfNeeded` end up with `String::format`, considering moving `String::format` there

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -395,11 +403,12 @@ private void emitGlobalGetter(String holderClass, String javaName,
...
         Class<?> type = Utils.carrierFor(decl.type());
-        appendLines(STR."""
-            public static \{type.getSimpleName()} \{javaName}() {
-                return \{holderClass}.SEGMENT.get(\{holderClass}.LAYOUT, 0L);
+        appendLines(String.format("""
+            public static %s %s() {
+                return %s.SEGMENT.get(%s.LAYOUT, 0L);
             }
-            """);
+            """, type.getSimpleName(), javaName, holderClass, holderClass));
```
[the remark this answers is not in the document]
mcimadamore: This might also help improving readability for messages containing lots of `%s` (as at least you can look at the number being referenced, and then count the arguments passed to the `format` method to find the one you want. Also, this allows to refactor code more robustly - e.g. you can pass _new_ trailing parameters to `format`, and maybe have them used _before_ the others:

```
jshell> MessageFormat.format("{0} {2} {0} {1}", "Hello", "World!", "Maurizio");
$7 ==> "Hello Maurizio Hello World!"
```

E.g. thanks to explicit numbers, the formatting is not tied to the order in which the arguments are passed to the `format` method.

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -558,32 +569,45 @@ private void emitConstant(Class<?> javaType, String constantName, Object value,
...
-                public static \{javaType.getSimpleName()} \{constantName}() {
-                    class Holder {
-                        static final \{javaType.getSimpleName()} \{constantName}
-                            = \{runtimeHelperName()}.LIBRARY_ARENA.allocateFrom("\{Utils.quote(Objects.toString(value))}");
-                    }
-                    return Holder.\{constantName};
-                }
-                """);
+            appendLines(String.format("""
+        public static %s %s() {
```
mcimadamore, line 573: Why the negative indentation here?

src/main/java/org/openjdk/jextract/impl/StructBuilder.java
```diff
@@ -282,70 +292,78 @@ private void emitFieldArraySetter(String javaName, Declaration.Variable varTree,
...
+             * The returned segment has address %s.address() + index * layout().byteSize()
              */
-            public static MemorySegment asSlice(MemorySegment \{arrayParam}, long index) {
-                return \{arrayParam}.asSlice(layout().byteSize() * index);
+            public static MemorySegment asSlice(MemorySegment %s, long index) {
+                return %s.asSlice(layout().byteSize() * index);
             }
-            """);
+            """,
+            arrayParam, arrayParam, arrayParam, arrayParam, arrayParam));
```
mcimadamore, line 325: In this case, using a `MessageFormat`-like approach is particularly convenient (as there would be just one param)

mcimadamore (commented):
This is a good like-for-like refactoring. I've left comments to improve readability of the code a bit. I think the comments are of two kinds:
* we could move String::format deeper in the call chain, so that call sites could look "cleaner"
* we could use a different way to format messages, more similar to `MessageFormat` where arguments have indices that can be used in the format message (so they can be repeated, reordered, etc.).

Addressing the latter comment is not strictly required as part of this PR, but I believe we might need to do something along these lines to restore at least some level of readability.

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
nizarbenalla: Would this work? 

Before
```
 appendIndentedLines(String.format("""
                private static final MethodHandle %s = %s.sliceHandle(%s);
                """,
                arrayHandleName, fieldLayoutName, path));
```
After 
```
void appendIndentedLines(String s, String... args) {
        sb.appendIndentedLines(String.format(s,args));
    }
```
and the call would look like 
```
appendIndentedLines("private static final MethodHandle %s = %s.sliceHandle(%s);"
                                , arrayHandleName, fieldLayoutName, path));
```
probably would create a new method rather than remove the old one

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
nizarbenalla: Fixed in [6adcab3](https://github.com/openjdk/jextract/pull/244/commits/6adcab3e23db3e03ee08a05889e23de31d088d2d), I think I didn't see them in the diff.

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
nizarbenalla: Fixed in [991ad7c](https://github.com/openjdk/jextract/pull/244/commits/991ad7c1a6f2c13aa9c7860cf654432e8bc70bac)

nizarbenalla:
Thanks for the comments!

JornVernee:
> we could use a different way to format messages, more similar to MessageFormat where arguments have indices that can be used in the format message (so they can be repeated, reordered, etc.).

Note that `String.format` also supports referencing arguments by index, e.g. `String.format("%2$s %1$s", "A", "B")` would produce `"B A"`, and the same argument can be referenced multiple times as well.

mcimadamore:
> > we could use a different way to format messages, more similar to MessageFormat where arguments have indices that can be used in the format message (so they can be repeated, reordered, etc.).
> 
> `String.format` also supports referencing arguments by index, e.g. `String.format("%2$s %1$s", "A", "B")` would produce `"B A"`, and the same argument can be referenced multiple times as well. I suggest we use that for the longer text blocks with multiple embedded values.

Good point, I never used this escape hatch :-)
I wonder though if readability of that is good enough. The problem with message format of course is need to escape `{` which are quite common in code...

So perhaps we can use the trick suggested by @JornVernee  in selected places and maybe tinker later (e.g. maybe come up with our own templating solution, if we really need that).

nizarbenalla:
I hope this approach is more readable.
I used `MessageFormat` sparsely, only when the same argument was references multiple times. As code fragments have a lot of curly braces, having to escape them each time it hurt readability.

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
```diff
@@ -107,7 +107,7 @@ final void classBegin() {
         if (superName != null) {
             extendsExpr = " extends " + superName;
         }
-        appendLines(String.format("%s %s %s%s {", modifiers, kind.kindName, className, extendsExpr));
+        appendLines("%s %s %s%s {", modifiers, kind.kindName, className, extendsExpr);
```
mcimadamore, line 110: no numbers here?

nizarbenalla: Fixed in [8507b05](https://github.com/openjdk/jextract/pull/244/commits/8507b0533a46d54cfbd8b040aeddcead3aa70e6d), thanks

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -435,11 +440,11 @@ private void emitGlobalSegmentSetter(String holderClass, String javaName,
...
         emitDocComment(varTree, docHeader);
-        appendLines(STR."""
-            public static void \{javaName}(MemorySegment varValue) {
-                MemorySegment.copy(varValue, 0L, \{holderClass}.SEGMENT, 0L, \{holderClass}.LAYOUT.byteSize());
+        appendLines("""
+            public static void %s(MemorySegment varValue) {
+                MemorySegment.copy(varValue, 0L, %s.SEGMENT, 0L, %s.LAYOUT.byteSize());
             }
-            """);
+            """, javaName, holderClass, holderClass);
```
JornVernee, line 447: I suggest using numbers here too, since the param is repeated

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -451,21 +456,22 @@ private void emitGlobalArrayGetter(String holderClass, IndexList indexList,
...
+                public static MemorySegment %1$s(%2$s) {
                     try {
-                        return (MemorySegment)\{holderClass}.HANDLE.invokeExact(\{holderClass}.SEGMENT, 0L, \{indexList.use()});
+                        return (MemorySegment)%3$s.HANDLE.invokeExact(%4$s.SEGMENT, 0L, %5$s);
                     } catch (Throwable ex$) {
                         throw new AssertionError("should not reach here", ex$);
                     }
                 }
-                """);
+                """, javaName, indexList.decl(), holderClass, holderClass, indexList.use());
```
JornVernee, line 467: Param is repeated here too, so could re-use `%3$s`

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -478,24 +484,24 @@ private void emitGlobalArraySetter(String holderClass, IndexList indexList,
...
-            appendLines(STR."""
-                public static void \{javaName}(\{indexList.decl()}, MemorySegment varValue) {
-                    MemorySegment.copy(varValue, 0L, \{javaName}(\{indexList.use()}), 0L, \{layoutString(elemType)}.byteSize());
-                }
-                """);
+            appendLines(MessageFormat.format("""
+        public static void {0}({1}, MemorySegment varValue) '{'
+            MemorySegment.copy(varValue, 0L, {0}({2}), 0L, {3}.byteSize());
+        '}'
+        """, javaName, indexList.decl(), indexList.use(), layoutString(elemType)));
```
JornVernee, line 491: Is this the only place where MessageFormat is used? Why just here?

(also, indentation was dropped here)

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -310,23 +315,23 @@ void emitFirstHeaderPreamble(List<Options.Library> libraries, boolean useSystemL
             appendBlankLine();
             appendIndentedLines("""
 
-                static {
-                """);
+            static {
+            """);
```
JornVernee, line 319: Please keep the indentation, here and in other places.

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -558,32 +565,45 @@ private void emitConstant(Class<?> javaType, String constantName, Object value,
...
                 }
-                """);
+                """,
+                javaType.getSimpleName(),
+                constantName,
+                javaType.getSimpleName(),
+                constantName,
+                runtimeHelperName(),
+                Utils.quote(Objects.toString(value)),
+                constantName);
```
JornVernee, line 583: Numbers would be nice here to avoid repetition of parameters.

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -558,32 +565,45 @@ private void emitConstant(Class<?> javaType, String constantName, Object value,
...
-            appendLines(STR."""
-                public static \{javaType.getSimpleName()} \{constantName}() {
-                    return \{constantName};
+            appendLines("""
+                private static final %s %s = %s;
+                """,
+                javaType.getSimpleName(),
+                constantName,
+                constantValue(javaType, value));
+                 emitDocComment(declaration);
```
JornVernee, line 591:
```suggestion
            emitDocComment(declaration);
```

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -558,32 +565,45 @@ private void emitConstant(Class<?> javaType, String constantName, Object value,
...
+            appendLines("""
+                public static %s %s() {
+                    return %s;
                 }
-                """);
+                """,
+                javaType.getSimpleName(),
+                constantName,
+                constantName);
```
JornVernee, line 599: Numbers here too I think.

src/main/java/org/openjdk/jextract/impl/StructBuilder.java
```diff
@@ -201,50 +202,51 @@ private void emitFieldSetter(String javaName, Declaration.Variable varTree, Stri
...
     private void emitSegmentSetter(String javaName, Declaration.Variable varTree, String offsetField, String layoutField) {
         appendBlankLine();
         emitFieldDocComment(varTree, "Setter for field:");
         String segmentParam = safeParameterName(kindName());
         String valueParam = safeParameterName("fieldValue");
-        appendIndentedLines(STR."""
-            public static void \{javaName}(MemorySegment \{segmentParam}, MemorySegment \{valueParam}) {
-                MemorySegment.copy(\{valueParam}, 0L, \{segmentParam}, \{offsetField}, \{layoutField}.byteSize());
+        appendIndentedLines("""
+           public static void %1$s(MemorySegment %2$s, MemorySegment %3$s) {
```
JornVernee, line 230:
```suggestion
            public static void %1$s(MemorySegment %2$s, MemorySegment %3$s) {
```

src/main/java/org/openjdk/jextract/impl/StructBuilder.java
```diff
@@ -282,70 +285,71 @@ private void emitFieldArraySetter(String javaName, Declaration.Variable varTree,
...
+             * The returned segment has address {0}.address() + index * layout().byteSize()
              */
-            public static MemorySegment asSlice(MemorySegment \{arrayParam}, long index) {
-                return \{arrayParam}.asSlice(layout().byteSize() * index);
-            }
-            """);
+            public static MemorySegment asSlice(MemorySegment {0}, long index) '{'
+                return {0}.asSlice(layout().byteSize() * index);
+            '}'
+            """, arrayParam));
```
JornVernee, line 315: This doesn't look right. The javadoc should be changed here.

src/main/java/org/openjdk/jextract/impl/StructBuilder.java
```diff
@@ -282,70 +285,71 @@ private void emitFieldArraySetter(String javaName, Declaration.Variable varTree,
...
     }
 
     private void emitAllocatorAllocate() {
         String allocatorParam = safeParameterName("allocator");
-        appendIndentedLines(STR."""
+        appendIndentedLines(MessageFormat.format("""
 
             /**
-             * Allocate a segment of size {@code layout().byteSize()} using {@code \{allocatorParam}}
+             * Allocate a segment of size '{'@code layout().byteSize()'}' using {0}
```
JornVernee, line 332: Same here, we need to `@code` tag still.

src/main/java/org/openjdk/jextract/impl/StructBuilder.java
mcimadamore: The quoting of the `{` also seems suspicious?

nizarbenalla: This was one of the couple of times where I used `MessageFormat`, so I had to escape the `{`

nizarbenalla: This was one of the couple of times where I used MessageFormat, so I had to escape the {

mcimadamore:
Overall, it seems to me that we need to decide on a policy on when to use numbering, as the current patch seems inconsistent. I see three options:
1. don't use numbering
2. only use numbering when there are repeated occurrences of same format argument
3. use numbering everywhere

I think (1) is where we started from. The current patch seems to be a mix between (2) and (3). In general, (2) might be enough, as it avoids repetition on very long format calls, while not making readability of simple cases overly complex. But I'm also ok with (3) if we prefer keeping same style for all calls.

src/main/java/org/openjdk/jextract/impl/StructBuilder.java
mcimadamore: I presume we can just drop the call to `Message::format` here and keep `appendIndentedLines`, right?

nizarbenalla: Sure, I will do that.
I will try to cover all the suggested fixes in one patch

JornVernee:
> Overall, it seems to me that we need to decide on a policy on when to use numbering, as the current patch seems inconsistent. I see three options:
> 
>     1. don't use numbering
> 
>     2. only use numbering when there are repeated occurrences of same format argument
> 
>     3. use numbering everywhere
> 
> 
> I think (1) is where we started from. The current patch seems to be a mix between (2) and (3). In general, (2) might be enough, as it avoids repetition on very long format calls, while not making readability of simple cases overly complex. But I'm also ok with (3) if we prefer keeping same style for all calls.

I think it makes sense to use numbering for repeated values as well as longer snippets (more than 1-2 lines). Probably for consistency it's better to use numbers everywhere then

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
```diff
@@ -132,6 +130,9 @@ final void decrAlign() {
     void appendLines(String s) {
         sb.appendLines(s);
     }
+    void appendLines(String s, String... args) {
+        sb.appendLines(String.format(s, (Object []) args));
```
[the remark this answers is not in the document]
nizarbenalla: Short answer is that I would get a warning
```
src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java:134: warning: non-varargs call of varargs method with inexact argument type for last parameter;
        sb.appendLines(String.format(s,  args));
                                         ^
  cast to Object for a varargs call
  cast to Object[] for a non-varargs call and to suppress this warning
1 warning
```
Without the cast, javac treats `args` as a simple argument of type `String []` while this is the signature for `String.format`
```
public static String format(String format,Object... args)
```

nizarbenalla:
I went with option 3 and simply added numbering everywhere.

JornVernee (approved):
Looks good, thanks.

nizarbenalla:
Thanks Jorn for the review, I will wait a bit and see if Maurizio wants any additional changes.

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
```diff
@@ -143,28 +144,32 @@ void appendIndentedLines(String s) {
...
             /**
-            \{!header.isEmpty() ? STR." * \{header}\n" : ""}\
+            %1$s
              * {@snippet lang=c :
-            \{declarationComment(decl)}
+            %2$s
              * }
              */
-            """);
+            """, !header.isEmpty() ? String.format(" * %1$s\n", header) : "", declarationComment(decl));
```
mcimadamore, line 172: optional comment: we could add a `format` method (which just calls String::format) and then implement `appendIndentedLines` on top. I guess what I wonder is if, at some point, it might make sense to create a more specialized format syntax for jextract use to make `%1$s` a bit more readable.

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
```diff
@@ -235,31 +240,32 @@ String indentString(int size) {
...
         return switch (primitiveType.kind()) {
-            case Bool -> STR."\{runtimeHelperName()}.C_BOOL";
-            case Char -> STR."\{runtimeHelperName()}.C_CHAR";
-            case Short -> alignIfNeeded(STR."\{runtimeHelperName()}.C_SHORT", 2, align);
-            case Int -> alignIfNeeded(STR."\{runtimeHelperName()}.C_INT", 4, align);
-            case Long -> alignIfNeeded(STR."\{runtimeHelperName()}.C_LONG", TypeImpl.IS_WINDOWS ? 4 : 8, align);
-            case LongLong -> alignIfNeeded(STR."\{runtimeHelperName()}.C_LONG_LONG", 8, align);
-            case Float -> alignIfNeeded(STR."\{runtimeHelperName()}.C_FLOAT", 4, align);
-            case Double -> alignIfNeeded(STR."\{runtimeHelperName()}.C_DOUBLE", 8, align);
+            case Bool -> String.format("%1$s.C_BOOL", runtimeHelperName());
```
mcimadamore, line 243: is this more readable than `runtimeHelperName() + ".C_BOOL"` ? I know this is all subjective, but we have a lot of "simple" concatenation which now is using `String::format`. Perhaps worth cleaning up in a separate PR.

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -166,95 +166,99 @@ private void emitFunctionWrapper(String javaName, String nativeName, boolean nee
...
+            appendLines("""
+            public static %1$s %2$s(%3$s) {
+                var mh$ = %4$s.HANDLE;
                 try {
                     if (TRACE_DOWNCALLS) {
-                        traceDowncall(\{traceArgList});
+                        traceDowncall(%5$s);
                     }
-                    \{returnWithCast}mh$.invokeExact(\{paramList});
+                    %6$s mh$.invokeExact(%7$s);
```
mcimadamore, line 219: extra space here?

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -409,11 +413,11 @@ private void emitGlobalSetter(String holderClass, String javaName,
...
         Class<?> type = Utils.carrierFor(decl.type());
-        appendLines(STR."""
-            public static void \{javaName}(\{type.getSimpleName()} varValue) {
-                \{holderClass}.SEGMENT.set(\{holderClass}.LAYOUT, 0L, varValue);
+        appendLines("""
+            public static void %1$s(%2$s varValue) {
+                %3$s.SEGMENT.set(%4$s.LAYOUT, 0L, varValue);
             }
-            """);
+            """, javaName, type.getSimpleName(), holderClass, holderClass);
```
mcimadamore, line 420: repeated `holderClass`

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -478,24 +483,24 @@ private void emitGlobalArraySetter(String holderClass, IndexList indexList,
...
         if (Utils.isStructOrUnion(elemType)) {
-            appendLines(STR."""
-                public static void \{javaName}(\{indexList.decl()}, MemorySegment varValue) {
-                    MemorySegment.copy(varValue, 0L, \{javaName}(\{indexList.use()}), 0L, \{layoutString(elemType)}.byteSize());
+            appendLines("""
+                public static void %1$s(%2$s, MemorySegment varValue) {
+                    MemorySegment.copy(varValue, 0L, %1$s(%3$s), 0L, %4$s.byteSize());
                 }
-                """);
+                 """, javaName, indexList.decl(), indexList.use(), layoutString(elemType));
```
mcimadamore, line 490: changed indentation?

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -504,51 +509,52 @@ private String emitVarHolderClass(Declaration.Variable var, String javaName) {
                     .collect(Collectors.joining(", "));
             Type elemType = Utils.typeOrElemType(varType);
             String accessHandle = Utils.isStructOrUnion(elemType) ?
-                    STR."public static final MethodHandle HANDLE = LAYOUT.sliceHandle(\{path});" :
-                    STR."public static final VarHandle HANDLE = LAYOUT.varHandle(\{path});\n";
-            String dimsString = dimensions.stream().map(d -> d.toString())
+                    "public static final MethodHandle HANDLE = LAYOUT.sliceHandle(" + path + ");" :
```
mcimadamore, line 512: I note that here we're using string concat

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
```diff
@@ -126,12 +126,16 @@ final void decrAlign() {
         sb.decrAlign();
     }
 
+    String format(String s, String... args) {
```
nizarbenalla, line 129: I added this method here, then I use it in both `appendLines` and `appendIndentedLines`

mcimadamore (approved):
Looks good. I'm not too sure about readability of some cases e.g. when we have `String.format("%s %s", a, b)` and add indices (even thought parameters are used in order. But (a) we can address this in a separate PR and (b) for bigger snippets one might argue that it's hard to keep rack of the position at which an argument appears, so having the number in there is handy (so my objection only really applies to the small format strings) and (c) maybe the best way to resolve that is to (as suggested otherwise) use some other kind of custom formatting.

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
```diff
@@ -240,8 +240,8 @@ String indentString(int size) {
 
     private String primitiveLayoutString(Primitive primitiveType, long align) {
         return switch (primitiveType.kind()) {
-            case Bool -> String.format("%1$s.C_BOOL", runtimeHelperName());
-            case Char -> String.format("%1$s.C_CHAR", runtimeHelperName());
+            case Bool -> runtimeHelperName() + ".C_BOOL";
```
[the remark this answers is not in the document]
mcimadamore: If you fix these, shouldn't you fix also the other cases in this switch, for consistency?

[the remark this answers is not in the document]
mcimadamore: Ah ok, alignIfNeeded assumes formatting... but, is that a self-inflicted wound?

nizarbenalla:
I'm not sure if this project is the best place but we can maybe experiment with different formatters?
If no more changes are needed I'll integrate

/integrate

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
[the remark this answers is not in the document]
nizarbenalla: Let me see. With it being used only 10 times in the codebase, I could rework it

PR #246, opened 2024-05-27
Remove duplicate copies of libclang on Linux 

mcimadamore:
On Linux, libclang.so is found in three forms:

* `libclang.so.13.0.0`, this is the real library
* `libclang.so.13`, this is a symlink that points to the above library
* `libclang.so`, this is a symlink that points to the above symlink

Because of this, and because Gradle lacks ability to distinguish between real filenames and symlinks, we end up (on Linux) with multiple versions of libclang, which significantly impact on the size of the final jextract artifact.

The solution is to add some special logic on Linux where:

* only `libclang.so.13.0.0` is copied from LLVM
* `libclang.so.13.0.0` is later renamed to `libclang.so`

This PR implement such changes both for gradle and make build (both have been tested locally).

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/246/head:pull/246` \
`$ git checkout pull/246`

`$ git checkout pull/246` \
`$ git pull https://git.openjdk.org/jextract.git pull/246/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 246`

`$ git pr show -t 246`

</details>

<a href="https://git.openjdk.org/jextract/pull/246.diff">https://git.openjdk.org/jextract/pull/246.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/246#issuecomment-2133572088)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -73,7 +73,7 @@ task copyLibClang(type: Sync) {
 
     // make sure we only pick up the "real" shared library file from LLVM installation
     // (e.g. exclude symlinks on Linux)
-    def clang_path_include = Os.isFamily(Os.FAMILY_UNIX) ?
+    def clang_path_include = (Os.isFamily(Os.FAMILY_UNIX) && !Os.isFamily(Os.FAMILY_MAC)) ?
             "libclang.so.${clang_version}" : "*clang*"
 
     from("${libclang_dir}") {
--- a/build.gradle
+++ b/build.gradle
@@ -71,11 +71,18 @@ jar {
 task copyLibClang(type: Sync) {
     into("$buildDir/jmod_inputs")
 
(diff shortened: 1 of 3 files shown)


JornVernee (approved):
LGTM. Might want to test on mac as well?

mcimadamore:
> LGTM. Might want to test on mac as well?

Yes. My understanding is that on macos, versioning with symlinks is not a thing. That said, I think there's an issue in the makefile: when we do the final `mv`, if the file doesn't exist, then `mv` will fail.

JornVernee (approved):
Looks good. GHA failure on Windows is due to libclang crash

PR #249, opened 2024-06-19
7903755: multi header support and special syntax for header file

sundararajana:
jextract currently supports only one header file. If the user wants to extract more than one header file, s/he has to create a containing header that includes multiple headers and jextract the containing header.

With the current change, jextract supports more than one header file in the command line.
When multiple headers are specified in command line, --header-class-name option is mandatory and it is enforced.

In addition to normal file names for header file names, a special syntax such as "<stdio.h>" is also supported. With this, user does not have to write long file name including directories such as /usr/include/stdio.h.

jextract generates multiple headers by generating a temporary header file. If the command line argument is of the form "<stdio.h>" then jextract generates "#include <stdio.h>" in the auto-generated containing header. If the header specified in of the normal/existing form (say foo.h), then containing header will have #include "foo.h" line.

---------
### Progress

### Issue
 * [CODETOOLS-7903755](https://bugs.openjdk.org/browse/CODETOOLS-7903755): multi header support and special syntax for header file (**Enhancement** - P3)

### Reviewers

### Contributors
 * Jorn Vernee `<jvernee@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/249/head:pull/249` \
`$ git checkout pull/249`

`$ git checkout pull/249` \
`$ git pull https://git.openjdk.org/jextract.git pull/249/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 249`

`$ git pr show -t 249`

</details>

<a href="https://git.openjdk.org/jextract/pull/249.diff">https://git.openjdk.org/jextract/pull/249.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/249#issuecomment-2178286132)


src/main/java/org/openjdk/jextract/JextractTool.java
```diff
@@ -82,13 +82,22 @@ private JextractTool(Logger logger) {
         this.logger = logger;
     }
 
-    private static Path generateTmpSource(List<Path> headers) {
-        assert headers.size() > 1;
+    private static boolean isSpecialHeaderName(String str) {
+        return str.startsWith("<") && str.endsWith(">");
+    }
+
+    private static Path generateTmpSource(List<String> headers) {
```
mcimadamore, line 89: This creates a temp file. It would be good to avoid this and maybe use clang APIs to parse in-memory files?

mcimadamore (commented):
Samples are much better - all the scripts are easier to understand.
I think we need to update the guide - as some of the command line option have changed:
* e.g. we now accept multiple headers
* headers can be provided with two different syntax (e.g. plain and "<>")
* if there's multiple headers, then the header class name option *must* be used

src/main/java/org/openjdk/jextract/JextractTool.java
sundararajana: we're using the same trick for macro re-parser. I'll try to revisit this in a future PR

doc/GUIDE.md
```diff
@@ -986,6 +986,21 @@ A complete list of all the supported command line options is given below:
 | `--include-[function,constant,struct,union,typedef,var]<String>` | Include a symbol of the given name and kind in the generated bindings. When one of these options is specified, any symbol that is not matched by any specified filters is omitted from the generated bindings. |
 | `--version`                                                  | print version information and exit |
 
+jextract accepts one or more header files. When multiple header files are specified,
```
mcimadamore, line 989:
```suggestion
Jextract accepts one or more header files. When multiple header files are specified,
```

doc/GUIDE.md
```diff
@@ -986,6 +986,21 @@ A complete list of all the supported command line options is given below:
...
+--header-class-name option is mandatory. Header files can be specified in two different ways.
```
mcimadamore, line 990:
```suggestion
the `--header-class-name` option is mandatory. Header files can be specified in two different ways.
```

doc/GUIDE.md
```diff
@@ -986,6 +986,21 @@ A complete list of all the supported command line options is given below:
...
+
+   1. simple header file name like "foo.h" or header file path like "bar/foo.h"
```
mcimadamore, line 992:
```suggestion
   1. Simple header file name like "foo.h" or header file path like "bar/foo.h"
```

doc/GUIDE.md
```diff
@@ -986,6 +986,21 @@ A complete list of all the supported command line options is given below:
...
+
+   2. Special header file path or file name like "<stdio.h>", "<GLUT/glut.h>".
+      With this syntax, the header full or relative path is not required. The usual C compiler
```
mcimadamore, line 995: I think we should say something like:
```
With this syntax, the header path is considered to be relative to one of the paths in the C compiler include path.
```

doc/GUIDE.md
```diff
@@ -986,6 +986,21 @@ A complete list of all the supported command line options is given below:
...
+      include path search is used in this case. This simplifies the extraction of header files
+      from standard include paths and include paths specified by -I options.
```
mcimadamore, line 997:
```suggestion
      from standard include paths and include paths specified by `-I` options.
```

doc/GUIDE.md
```diff
@@ -986,6 +986,21 @@ A complete list of all the supported command line options is given below:
 | `--include-[function,constant,struct,union,typedef,var]<String>` | Include a symbol of the given name and kind in the generated bindings. When one of these options is specified, any symbol that is not matched by any specified filters is omitted from the generated bindings. |
 | `--version`                                                  | print version information and exit |
 
+jextract accepts one or more header files. When multiple header files are specified,
+--header-class-name option is mandatory. Header files can be specified in two different ways.
+
+   1. simple header file name like "foo.h" or header file path like "bar/foo.h"
+
+   2. Special header file path or file name like "<stdio.h>", "<GLUT/glut.h>".
```
mcimadamore, line 994:
```suggestion
   2. Special header file path or file name like `<stdio.h>`, `<GLUT/glut.h>`.
```

mcimadamore (commented):
Note: the jextract guide starts off with an example like this:

```
$ jextract \
  --include-dir /path/to/mylib/include \
  --output src \
  --target-package org.jextract.mylib \
  --library mylib \
  /path/to/mylib/include/mylib.h
```
I think we should either update the example or (maybe preferred, to keep things simple) add a simple para where we say that the command line can be simplified by having jextract look for headers in include paths (link to option section).

doc/GUIDE.md
```diff
@@ -58,10 +58,12 @@ In this command:
 
 Note that specifying the wrong header file to jextract may result in errors during parsing.
 Please consult the documentation of the library in question about which header file
-should be included. This is also the header file that should be passed to jextract. If
-a library has multiple main header files, they can be passed to jextract by creating a new
-header file which `#include`s these header files, and then this new header file can be
-passed to jextract.
+should be included. This is also the header file that should be passed to jextract.
+
+If a library has multiple main header files, they can be passed to jextract in the command line.
```
JornVernee, line 63:
```suggestion
If a library has multiple main header files, they can be passed to jextract on the command line.
```

nlisker:
Maybe the command line options table entry for `--header-class-name <name>` should be updated regarding the mandatory use of it. Currently it says "If this option is not specified", but this is not always valid anymore.

doc/GUIDE.md
```diff
@@ -986,6 +988,21 @@ A complete list of all the supported command line options is given below:
...
+the `--header-class-name` option is mandatory. Header files can be specified in two different ways:
+
+   1. Simple header file name like `foo.h` or header file path like `bar/foo.h`
+
+   2. Special header file path or file name like `<stdio.h>`, `<GLUT/glut.h>`.
+      With this syntax, the header path is considered to be relative to one of the paths
+      in the C compiler include path. This simplifies the extraction of header files
+      from standard include paths and include paths specified by `-I` options.
+
+      Note that '>' and '<' are special characters in OS Shells and therefore those
```
mcimadamore, line 1001:
```suggestion
      Note that `>` and `<` are special characters in OS Shells and therefore those
```

doc/GUIDE.md
```diff
@@ -986,6 +988,21 @@ A complete list of all the supported command line options is given below:
 | `--include-[function,constant,struct,union,typedef,var]<String>` | Include a symbol of the given name and kind in the generated bindings. When one of these options is specified, any symbol that is not matched by any specified filters is omitted from the generated bindings. |
 | `--version`                                                  | print version information and exit |
 
+Jextract accepts one or more header files. When multiple header files are specified,
```
JornVernee, line 991: I suggest adding a header above this section to separate it from the table (like we currently have for the section on clang arguments below)

sundararajana: Hmm.. not sure what would be the name. Perhaps move this para above the table?

JornVernee (approved):

doc/GUIDE.md
JornVernee: Discussed offline: let's keep it like this

mcimadamore (approved):

PR #251, opened 2024-06-25
7903758: jextract should avoid temporary file creation

sundararajana:
jextract creates a temporary file to support multiple headers and special header syntax <foo.h>. This patch avoids that temporary file creation by using clang's in-memory parsing API clang_createTranslationUnitFromSourceFile

---------
### Progress

### Issue
 * [CODETOOLS-7903758](https://bugs.openjdk.org/browse/CODETOOLS-7903758): jextract should avoid temporary file creation (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/251/head:pull/251` \
`$ git checkout pull/251`

`$ git checkout pull/251` \
`$ git pull https://git.openjdk.org/jextract.git pull/251/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 251`

`$ git pr show -t 251`

</details>

<a href="https://git.openjdk.org/jextract/pull/251.diff">https://git.openjdk.org/jextract/pull/251.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/251#issuecomment-2188682229)

### DIFF
--- a/src/main/java/org/openjdk/jextract/clang/Index.java
+++ b/src/main/java/org/openjdk/jextract/clang/Index.java
@@ -62,27 +62,43 @@ public class Index extends ClangDisposable {
         private final ErrorCode code;
 
         public ParsingFailedException(Path srcFile, ErrorCode code) {
-            super("Failed to parse " + srcFile.toAbsolutePath().toString() + ": " + code);
-            this.srcFile = srcFile.toAbsolutePath().toString();
+            this(srcFile.toAbsolutePath().toString(), code);
+        }
+
+        public ParsingFailedException(String srcFile, ErrorCode code) {
+            super("Failed to parse " + srcFile + ": " + code);
+            this.srcFile = srcFile;
             this.code = code;
         }
     }
 
-    public TranslationUnit parseTU(String file, Consumer<Diagnostic> dh, int options, String... args)
-            throws ParsingFailedException {
+    private TranslationUnit parseTUImpl(String file, String content,
+                Consumer<Diagnostic> dh, int options, String... args)
+                throws ParsingFailedException {
         try (Arena arena = Arena.ofConfined()) {
-            MemorySegment src = arena.allocateFrom(file);
+            MemorySegment fileSeg = arena.allocateFrom(file);
+            MemorySegment contentSeg = content == null ? null : arena.allocateFrom(content);
             MemorySegment cargs = args.length == 0 ? null : arena.allocate(C_POINTER, args.length);
             for (int i = 0 ; i < args.length ; i++) {
                 cargs.set(C_POINTER, i * C_POINTER.byteSize(), arena.allocateFrom(args[i]));
             }
+
+            MemorySegment unsavedFile = contentSeg == null ?
+                null : CXUnsavedFile.allocate(arena);
+            if (unsavedFile != null) {
+                CXUnsavedFile.Filename(unsavedFile, fileSeg);
+                CXUnsavedFile.Contents(unsavedFile, contentSeg);
+                CXUnsavedFile.Length(unsavedFile, content.length());
+            }
+
             MemorySegment outAddress = arena.allocate(C_POINTER);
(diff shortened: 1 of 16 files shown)


src/main/java/org/openjdk/jextract/clang/Index.java
```diff
@@ -98,6 +99,36 @@ public TranslationUnit parseTU(String file, Consumer<Diagnostic> dh, int options
...
+                    cargs == null ? MemorySegment.NULL : cargs,
+                    1,
+                    unsavedFile);
+            TranslationUnit rv = new TranslationUnit(tu);
+            // even if we failed to parse, we might still have diagnostics
+            rv.processDiagnostics(dh);
+
+            return rv;
+        }
+    }
```
JornVernee, line 130: Can't we use `clang_parseTranslationUnit2`? It also supports unsaved files, but also supports returning error codes. I suppose `clang_createTranslationUnitFromSourceFile` return `NULL` if parsing fails.

JornVernee: I suggest extending the existing `parseTU` to also accept a list of `Index.UnsavedFile`

src/main/java/org/openjdk/jextract/impl/Parser.java
```diff
@@ -60,49 +104,21 @@ public Declaration.Scoped parse(Path path, Collection<String> args) {
...
-                            decls.add(decl);
-                        }
-                    }
-                } else if (isMacro(c) && src.path() != null) {
-                    SourceRange range = c.getExtent();
-                    String[] tokens = c.getTranslationUnit().tokens(range);
-                    Optional<Declaration.Constant> constant = macroParser.parseConstant(c, c.spelling(), tokens);
-                    if (constant.isPresent()) {
-                        decls.add(constant.get());
+    public Declaration.Scoped parse(String name, String content, Collection<String> args) {
```
JornVernee, line 111: The only use of this method seems to be in `JextractTool`, so I think the old version can just be removed?

sundararajana: Will remove it.

src/main/java/org/openjdk/jextract/clang/Index.java
```diff
@@ -61,27 +62,43 @@ public static class ParsingFailedException extends RuntimeException {
         private final ErrorCode code;
 
         public ParsingFailedException(Path srcFile, ErrorCode code) {
-            super("Failed to parse " + srcFile.toAbsolutePath().toString() + ": " + code);
-            this.srcFile = srcFile.toAbsolutePath().toString();
+            this(srcFile.toAbsolutePath().toString(), code);
+        }
+
+        public ParsingFailedException(String srcFile, ErrorCode code) {
```
JornVernee, line 68: Where is this new constructor being used?

src/main/java/org/openjdk/jextract/clang/Index.java
```diff
@@ -98,6 +115,17 @@ public TranslationUnit parseTU(String file, Consumer<Diagnostic> dh, int options
         }
     }
 
+
+    public TranslationUnit parseTU(String file, Consumer<Diagnostic> dh, int options, String... args)
```
JornVernee, line 119: Can't the methods here that _don't_ take a file content be cleaned up as well now?

sundararajana: It was used in macro parser. But with changes in macro parse implementation, these may not be. Will remove.

src/main/java/org/openjdk/jextract/clang/Index.java
sundararajana: will check and remove one of the constructors

JornVernee (approved):

PR #253, opened 2024-07-16
7903776: jextract generates uncompilable code when C identifier with the name from "java.lang" classes is used

sundararajana:
Unqualified names are mangled by NameMangler. But it was complete (didn't include "Boolean"). For uniformity & future proof mangling all type names from "java.lang" package.

---------
### Progress

### Issue
 * [CODETOOLS-7903776](https://bugs.openjdk.org/browse/CODETOOLS-7903776): jextract generates uncompilable code when C identifier with the name from "java.lang" classes is used (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/253/head:pull/253` \
`$ git checkout pull/253`

`$ git checkout pull/253` \
`$ git pull https://git.openjdk.org/jextract.git pull/253/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 253`

`$ git pr show -t 253`

</details>

<a href="https://git.openjdk.org/jextract/pull/253.diff">https://git.openjdk.org/jextract/pull/253.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/253#issuecomment-2230921781)

### DIFF
--- a/src/main/java/org/openjdk/jextract/impl/NameMangler.java
+++ b/src/main/java/org/openjdk/jextract/impl/NameMangler.java
@@ -319,6 +319,11 @@ public final class NameMangler implements Declaration.Visitor<Void, Declaration>
     }
 
     private static boolean isJavaTypeName(String name) {
+        // rule out common case of lower-case starting identifier.
+        if (name.length() == 0 || Character.isLowerCase(name.charAt(0))) {
+            return false;
+        }
+
         // Java types that are used unqualified in the generated code
         return switch (name) {
             case "String", "Struct", "MethodHandle",
@@ -327,7 +332,16 @@ public final class NameMangler implements Declaration.Visitor<Void, Declaration>
                 "MemoryLayout",
                 "Arena", "NativeArena", "MemorySegment", "ValueLayout"
                     -> true;
-            default -> false;
+            default -> {
+                try {
+                    // java.lang package is auto-imported. Check if we've clash
+                    // of the given identifier with any of the java.lang.* types.
+                    Class.forName("java.lang." + name);
+                    yield true;
+                } catch (Exception ignored) {
+                }
+                yield false;
+            }
         };
     }
 }
--- a/src/main/java/org/openjdk/jextract/impl/NameMangler.java
+++ b/src/main/java/org/openjdk/jextract/impl/NameMangler.java
@@ -35,12 +35,14 @@ import org.openjdk.jextract.impl.DeclarationImpl.AnonymousStruct;
 import org.openjdk.jextract.impl.DeclarationImpl.JavaFunctionalInterfaceName;
 import org.openjdk.jextract.impl.DeclarationImpl.JavaName;
 
+import java.lang.reflect.Modifier;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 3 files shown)


src/main/java/org/openjdk/jextract/impl/NameMangler.java
```diff
@@ -319,6 +319,11 @@ private static boolean isRestrictedTypeName(String name) {
...
 
     private static boolean isJavaTypeName(String name) {
+        // rule out common case of lower-case starting identifier.
+        if (name.length() == 0 || Character.isLowerCase(name.charAt(0))) {
+            return false;
+        }
+
         // Java types that are used unqualified in the generated code
         return switch (name) {
             case "String", "Struct", "MethodHandle",
```
mcimadamore, line 329: should `String` be dropped from this list? Also, there seem to be other names that are out of sync, such as `NativeArena` and `Struct` (I wonder if we just want to repeat your trick for `java.lang.foreign`) and some are missing (e.g. `MethodHandle` but  not `VarHandle` ?)

src/main/java/org/openjdk/jextract/impl/NameMangler.java
```diff
@@ -327,7 +332,16 @@ private static boolean isJavaTypeName(String name) {
                 "MemoryLayout",
                 "Arena", "NativeArena", "MemorySegment", "ValueLayout"
                     -> true;
-            default -> false;
+            default -> {
+                try {
+                    // java.lang package is auto-imported. Check if we've clash
+                    // of the given identifier with any of the java.lang.* types.
+                    Class.forName("java.lang." + name);
```
liach, line 339: 1. Should we cache the results?
2. This detects package-private classes in java.lang too, which might not be desired. Maybe `yield clazz.accessFlags().contains(AccessFlag.PUBLIC)`

src/main/java/org/openjdk/jextract/impl/NameMangler.java
sundararajana: VarHandle and MethodHandle both are already included. Perhaps we can do same trick for java.lang.foreign.

src/main/java/org/openjdk/jextract/impl/NameMangler.java
sundararajana: Will fix

src/main/java/org/openjdk/jextract/impl/NameMangler.java
```diff
@@ -318,30 +320,46 @@ private static boolean isRestrictedTypeName(String name) {
...
+
+    private static boolean checkTypeInPackage(String pkgName, String typeName) {
+        try {
+            int mods = Class.forName(pkgName + "." + typeName).getModifiers();
+            if (Modifier.isPublic(mods)) {
+                WELL_KNOWN_JAVA_TYPES.add(typeName);
+                return true;
             }
-        };
+        } catch (Exception ignored) {
```
mcimadamore, line 361: Should we report something when using the verbose debug option?

mcimadamore (approved):

src/main/java/org/openjdk/jextract/impl/NameMangler.java
sundararajana: Hmm. Not sure. May be bit noisy I think as exceptions will be thrown for many uppercase starting C identifiers (enum constants, #defines)

PR #257, opened 2024-08-16
7903788: preparation towards json configuration for jextract tool

sundararajana:
* adding json skara code into org.openjdk.jextract.json.parser package. Minor changes done - pretty printing, JSONValue is made sealed and access changes (private final fields).
* org.openjdk.jextract.json package for JSON interface to the rest of the jextract tool. Main interface is org.openjdk.jextract.json.JSON. parse and toString methods to convert record into JSON String and vice versa.
* Other classes are public mainly for testing purposes. build.gradle adjusted with the required --add-exports clauses so that jextract module does not have to export the new json packages.

---------
### Progress

### Issue
 * [CODETOOLS-7903788](https://bugs.openjdk.org/browse/CODETOOLS-7903788): preparation towards json configuration for jextract tool (**Enhancement** - P4)

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - **Reviewer**) ⚠️ Review applies to [13cc5106](https://git.openjdk.org/jextract/pull/257/files/13cc5106ba6f58d9c8ed5643115bdf23647f3013)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/257/head:pull/257` \
`$ git checkout pull/257`

`$ git checkout pull/257` \
`$ git pull https://git.openjdk.org/jextract.git pull/257/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 257`

`$ git pr show -t 257`

</details>

<a href="https://git.openjdk.org/jextract/pull/257.diff">https://git.openjdk.org/jextract/pull/257.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/257#issuecomment-2293369025)


build.gradle
```diff
@@ -59,6 +59,7 @@ repositories {
 compileJava {
     options.release = 22
     options.compilerArgs << "--enable-preview"
+    options.compilerArgs << "-Xlint:unchecked"
```
mcimadamore, line 62: Are there many of these? Would be preferrable to fix them in place with `@SuppressWarnings`

src/main/java/org/openjdk/jextract/json/parser/JWCC.java
```diff
@@ -0,0 +1,38 @@
...
+package org.openjdk.jextract.json.parser;
+
+/**
+ * JWCC is JSON With Commas and Comments. In addition to supporting all of JSON
+ * JWCC also supports trailing commas and comments. Comments can be either
+ * until single-line or multi-line.
+ *
+ * Comments are stripped and are not present in the parsed result.
+ */
+public final class JWCC {
```
mcimadamore, line 32: This class (and related test) don't seem very useful. I think we can just drop them

src/main/java/org/openjdk/jextract/json/parser/JSONArray.java
```diff
@@ -0,0 +1,182 @@
...
+    public boolean isArray() {
+        return true;
+    }
+
+    @Override
+    public JSONArray asArray() {
+        return this;
+    }
+
+    public JSONArray set(int i, boolean value) {
```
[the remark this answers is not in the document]
mcimadamore: I also note that all the other classes seem to be non-mutable...

src/main/java/org/openjdk/jextract/json/JSONObjects.java
```diff
@@ -0,0 +1,826 @@
...
+        Set<Object> set = new LinkedHashSet<>();
+        final int length = ja.size();
+        for (int i = 0; i < length; i++) {
+             set.add(toObject(ja.get(i), compType));
+        }
+        return set;
+    }
+
+    private static Object convertArray(JSONArray ja, Type targetType) {
+        if (targetType instanceof Class cls) {
```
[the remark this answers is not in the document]
mcimadamore: What's the advantage of having the arrayConverters? In the end they will still (a) create a new array and (b) loop over the contents. Wouldn't it make more sense to define all the array logic in here, remove `arrayConverters` and then just redefine all the implementations of the factories in `JSONArrays` as shortcuts to this method?

src/main/java/org/openjdk/jextract/json/JSONObjects.java
```diff
@@ -0,0 +1,826 @@
...
+
+    static {
+        addDecimalConverter(double.class, JSONDecimal::asDouble);
+        addDecimalConverter(Double.class, JSONDecimal::asDouble);
+        addDecimalConverter(Number.class, JSONDecimal::asDouble);
+        addDecimalConverter(Object.class, JSONDecimal::asDouble);
+        addDecimalConverter(float.class, JSONDecimal::asFloat);
+        addDecimalConverter(Float.class, JSONDecimal::asFloat);
+        addDecimalConverter(OptionalDouble.class, jd -> OptionalDouble.of(jd.asDouble()));
+        addDecimalConverter(BigDecimal.class, jd -> BigDecimal.valueOf(jd.asDouble()));
```
mcimadamore, line 678: Not sure we'll ever use this one TBH?

src/main/java/org/openjdk/jextract/json/JSONObjects.java
```diff
@@ -0,0 +1,826 @@
...
+    static {
+        addNumberConverter(byte.class, JSONNumber::asByte);
+        addNumberConverter(Byte.class, JSONNumber::asByte);
+        addNumberConverter(short.class, JSONNumber::asShort);
+        addNumberConverter(Short.class, JSONNumber::asShort);
+        addNumberConverter(int.class, JSONNumber::asInt);
+        addNumberConverter(Integer.class, JSONNumber::asInt);
+        addNumberConverter(long.class, JSONNumber::asLong);
+        addNumberConverter(Long.class, JSONNumber::asLong);
+        addNumberConverter(OptionalInt.class, jn -> OptionalInt.of(jn.asInt()));
```
mcimadamore, line 638: These last four seem odd

src/main/java/org/openjdk/jextract/json/JSONObjects.java
```diff
@@ -0,0 +1,826 @@
...
+    }
+
+    private static Type getSecondTypeArgument(ParameterizedType pt) {
+        return getNthTypeArgument(pt, 1);
+    }
+
+    private static Type getNthTypeArgument(ParameterizedType pt, int n) {
+        Type[] typeArgs = pt.getActualTypeArguments();
+        Type nthType = typeArgs.length > n ? typeArgs[n] : Object.class;
+        if (nthType instanceof WildcardType wt) {
```
mcimadamore, line 362: This logic is not correct. E.g. if something is `List<? super Integer>` it is not correct to assume that the type argument is `String` (and e.g. create a `Integer[]` as a result). Because `List<? extends Number>` can have `Number` elements which won't go in an `Integer[]`.

In general, I think it might be appropriate to dial down the level of "smartness" and try to provide something that is well-behaved, and good enough for what we need. We don't need to solve the general problem of mapping json to Java, we just need a mapping that is good enough for jextract to use.

So... I think we'd be ok if wildcards won't be accepted by the mapper.

src/main/java/org/openjdk/jextract/json/JSONObjects.java
```diff
@@ -0,0 +1,826 @@
...
+
+        try {
+            return cls.getConstructor(paramTypes);
+        } catch (ReflectiveOperationException roe) {
+            throw new RuntimeException("cannot find canonical constructor", roe);
+        }
+    }
+
+    // get the Record component value as a JSONValue
+    private static JSONValue getRecordProperty(Record record, RecordComponent rc) {
```
mcimadamore, line 392: getRecordComponentValue seems a better name here?

src/main/java/org/openjdk/jextract/json/JSONObjects.java
```diff
@@ -0,0 +1,826 @@
...
+    // JSONObject converter helpers
+
+    // exists only localized SuppressWarnings
+    @SuppressWarnings("unchecked")
+    private static Record convertToRecord(JSONObject jo, Class<?> cls) {
+        assert cls.isRecord();
+        return toRecord(jo, (Class<? extends Record>)cls);
+    }
+
+    private static Map<?,?> convertToMap(JSONObject jo, Type keyType, Type valueType) {
```
mcimadamore, line 418: This seems overly general - isn't a JSON object always a map from String to something else?

build.gradle
sundararajana: It was left over from debugging. Added SuppressWarnings as needed. Will remove this.

src/main/java/org/openjdk/jextract/json/parser/JWCC.java
sundararajana: That means we'll not support extensions like comments and trailing comma (optionally). I'll remove that JSON.parse method then.

test/testng/org/openjdk/jextract/test/json/parser/JSONParserTests.java
```diff
@@ -0,0 +1,622 @@
...
+
+        var parsed = parser.parse(s);
+        assertEquals(parsed.get("inner").get("a").asInt(), 1);
+        assertEquals(parsed.get("inner").get("b").asInt(), 2);
+        assertEquals(parsed.get("inner").get("c").asInt(), 3);
+    }
+
+    @Test
+    void testLargerJSONText() {
+        var text = "{\n" +
```
mcimadamore, line 205: Would it make sense, for this or other big chunks of JSON, to do a roundtrip test where we parse the text, we get a JSONValue, then we obtain the string representation from it, and we compare it against the text?

src/main/java/org/openjdk/jextract/json/JSONObjects.java
sundararajana: Will rename.

src/main/java/org/openjdk/jextract/json/JSONObjects.java
sundararajana: Okay. I'll remove Wildcard types.

src/main/java/org/openjdk/jextract/json/parser/JSONArray.java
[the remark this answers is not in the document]
sundararajana: Trying to avoid large scale changes to skara json code... Changes are all minor (like package renaming, sealing, access change etc. only).

src/main/java/org/openjdk/jextract/json/parser/JSONNumber.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return Long.toString(value);
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
```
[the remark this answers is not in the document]
sundararajana: As I said, I didn't touch skara code much except for cosmetic changes. Skara's json mostly uses JDK 8 level code.

src/main/java/org/openjdk/jextract/json/JSONObjects.java
sundararajana: Removed BigDecimal. float exists only for completeness of 'all primitive' types.

src/main/java/org/openjdk/jextract/json/JSONObjects.java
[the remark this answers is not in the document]
sundararajana: > What about j.l.reflect.GenericArrayType? That seems like a valid target for this?

I though of GenericArrayType. Then realised that I've walk to the bottom type and find the eraser type.. and then build array-of-array.. (up to whatever dimension. Say List<String>[][][]). Not sure if that is worth it.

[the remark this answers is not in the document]
sundararajana: > What's the advantage of having the arrayConverters? In the end they will still (a) create a new array and (b) loop over the contents. Wouldn't it make more sense to define all the array logic in here, remove `arrayConverters` and then just redefine all the implementations of the factories in `JSONArrays` as shortcuts to this method?

Wouldn't that involve comparing Class object with a series of if-else-if for array component type? if (type == boolean.class).. else if (type == char.class)... unless I miss something.

src/main/java/org/openjdk/jextract/json/JSONObjects.java
```diff
@@ -0,0 +1,826 @@
...
+    }
+
+    static {
+        addDecimalConverter(double.class, JSONDecimal::asDouble);
+        addDecimalConverter(Double.class, JSONDecimal::asDouble);
+        addDecimalConverter(Number.class, JSONDecimal::asDouble);
+        addDecimalConverter(Object.class, JSONDecimal::asDouble);
+        addDecimalConverter(float.class, JSONDecimal::asFloat);
+        addDecimalConverter(Float.class, JSONDecimal::asFloat);
+        addDecimalConverter(OptionalDouble.class, jd -> OptionalDouble.of(jd.asDouble()));
```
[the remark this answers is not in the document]
sundararajana: You mean the Optional part? Thats to have property that can be dropped. float exists only for completion (of all primitive types).

src/main/java/org/openjdk/jextract/json/JSONObjects.java
sundararajana: You mean the OptionalInt part? Thats to have property that can be dropped. long vs int targets are supported (JSONNumber is always long - is converted to smaller ints by bounds check)

src/main/java/org/openjdk/jextract/json/JSONObjects.java
sundararajana: Keys can be anything convertible from String (like enum for eg).  Yes, there are not that many, but.. 

record MyRecord(Map<SymbolType, <valuetype>> ....)

test/testng/org/openjdk/jextract/test/json/parser/JSONParserTests.java
sundararajana: This test comes from skara repo. Other than package naming nothing is changed. Should I add a round trip test there itself or in another new test?

src/main/java/org/openjdk/jextract/json/JSONObjects.java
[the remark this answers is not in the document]
mcimadamore: Ah ok, you want to avoid the `if`. Fine.

mcimadamore (changes requested):
The parser addition and tests look ok.
I'm not overall convinced on the need for the mapper classes, and I think we should proceed w/o it for now, and see how bad it is.

build.gradle
mcimadamore: Lines are still there in the latest revision?

mcimadamore (approved):
Looks good (pending build changes)

PR #262, opened 2024-12-06
7903878: jextract should derive struct field offsets from struct layouts

nizarbenalla:
Please read the JBS issue for more details.

Currently jextract emits hardwired offsets, it would be better if the offsets could be computed dynamically from the struct layouts.

Passes all tests, unsure if this patch requires new tests.
Note: I used `fieldElementPaths(field.name())))` instead of the `javaName` to deal with nested structs.

It seems like GitHub actions are broken for Ubuntu? will look into this next week if necessary.


mcimadamore (approved):
Looks good. In principle we could save the construction of `fieldElementPaths`, as we use that both in the offset and in the layout accessor (e.g. `emitOffsetFieldDecl` and `emitLayoutFieldDecl`) but I'm not sure how much that matters in practice, as I'd expect most layout paths to be a simple call to `groupElement(String)`.

nizarbenalla:
On the same platforms, the offset are same the before/after the patch.

/integrate

PR #268, opened 2024-12-27
7903731: Jextract should support macOS frameworks

nizarbenalla:
Read the JBS issue for more detail.

This patch adds two new mac specific options, to make it easier to use frameworks and remove the need of a `compile_flags.txt` file. An exception is thrown if those options are used from an other platform, the error message is inspired from `jpackage`

### DIFF
--- a/doc/GUIDE.md
+++ b/doc/GUIDE.md
@@ -988,6 +988,14 @@ A complete list of all the supported command line options is given below:
 | `--include-[function,constant,struct,union,typedef,var]<String>` | Include a symbol of the given name and kind in the generated bindings. When one of these options is specified, any symbol that is not matched by any specified filters is omitted from the generated bindings. |
 | `--version`                                                  | print version information and exit |
 
+**macOS platform options (available only when running on macOS):**
+
+| Option                       | Meaning                                                                                                                                                                                                                                 |
+|:-----------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `--mac-framework-dir <dir>`  | specify the framework directory dir include files <br/>defaults to the current Mac OS X SDK dir<br/> This removes the need of having a compile_flags.txt with the required `-framework XYZ` options in the folder where jextract is ran |
+| `-f, framework <framekwork>` | specify the name of the library, path will be expanded to that of the framework folder                                                                                                                                                  |
+
+
 Jextract accepts one or more header files. When multiple header files are specified,
 the `--header-class-name` option is mandatory. Header files can be specified in two different ways:
 
--- a/src/main/java/org/openjdk/jextract/JextractTool.java
+++ b/src/main/java/org/openjdk/jextract/JextractTool.java
@@ -41,17 +41,11 @@ import org.openjdk.jextract.impl.UnsupportedFilter;
 
 import java.io.IOException;
 import java.io.PrintWriter;
-import java.io.UncheckedIOException;
 import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.List;
-import java.util.Optional;
+import java.util.*;
 import java.util.spi.ToolProvider;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -65,6 +59,8 @@ import java.util.stream.Stream;
 public final class JextractTool {
 
     public static final boolean DEBUG = Boolean.getBoolean("jextract.debug");
+    private static final boolean isMacOSX =
+            System.getProperty("os.name", "unknown").equals("Mac OS X");
 
     // error codes
     private static final int SUCCESS       = 0;
@@ -178,6 +174,9 @@ public final class JextractTool {
 
     private int printHelp(int exitCode) {
         logger.info("jextract.usage");
+        if (isMacOSX){
+            logger.info("jextract.usage.mac");
+        }
         return exitCode;
     }
 
@@ -243,6 +242,17 @@ public final class JextractTool {
         // option name to corresponding OptionSpec mapping
         private Map<String, OptionSpec> optionSpecs = new HashMap<>();
 
+        private static HashSet <String> platformOptions = new HashSet<>();
+
+        void addPlatformOption(String name) {
+            platformOptions.add(name);
+        }
+
+        void addPlatformOption(String name, List<String> aliases) {
+            platformOptions.add(name);
+            platformOptions.addAll(aliases);
+        }
+
         void accepts(String name, String help, boolean argRequired) {
             accepts(name, List.of(), help, argRequired);
         }
@@ -280,6 +290,10 @@ public final class JextractTool {
             return str.substring(2);
         }
 
+        static boolean checkIfSupported( String option) {
+            return platformOptions.contains(option);
+        }
+
         OptionSet parse(String[] args) {
             Map<String, List<String>> options = new HashMap<>();
             List<String> nonOptionArgs = new ArrayList<>();
@@ -293,14 +307,19 @@ public final class JextractTool {
                    // check for single char option followed
                    // by option value without whitespace in between.
                    // Examples: -lclang, -DFOO
-                   if (spec == null ) {
+                   if (spec == null) {
                        spec = isSingleCharOptionWithArg(arg) ? optionSpecs.get(singleCharOption(arg)) : null;
                        // we have a matching single char option and that requires argument
                        if (spec != null && spec.argRequired()) {
                            argValue = singleCharOptionArg(arg);
                        } else {
                            // single char option special handling also failed. give up.
-                           throw new OptionException("invalid option: " + arg);
+                           if (checkIfSupported(arg)) {
+                               var message = String.format("Error: Option [%s] is not valid on this platform", arg);
+                               throw new OptionException(message);
+                           } else {
+                               throw new OptionException("invalid option: " + arg);
+                           }
                        }
                    }
                    // handle argument associated with the current option, if any
@@ -365,6 +384,14 @@ public final class JextractTool {
         parser.accepts("-t", List.of("--target-package"), "help.t", true);
         parser.accepts("--version", "help.version", false);
 
+        if(isMacOSX) {
+            parser.accepts("--mac-framework-dir", "help.mac.framework", true);
+            parser.accepts("-f", "help.framework.library.path", true);
+        }
+
+        parser.addPlatformOption("--mac-framework-dir");
+        parser.addPlatformOption("-f");
+
         OptionSet optionSet;
         try {
             optionSet = parser.parse(args);
@@ -411,9 +438,8 @@ public final class JextractTool {
             builder.addClangArg("-I" + builtinInc);
         }
 
-        inferPlatformIncludePath().ifPresent(platformPath -> {
-            builder.addClangArg("-I" + platformPath);
-        });
+
+        inferPlatformIncludePath().ifPresent(platformPath -> builder.addClangArg("-I" + platformPath));
 
         String jextractHeaderPath = System.getProperty("jextract.header.path");
         if (jextractHeaderPath != null) {
@@ -442,26 +468,18 @@ public final class JextractTool {
             builder.setUseSystemLoadLibrary(true);
         }
 
-        boolean librariesSpecified = optionSet.has("-l");
-        if (librariesSpecified) {
-            for (String lib : optionSet.valuesOf("-l")) {
-                try {
-                    Library library = Options.Library.parse(lib);
-                    Path libPath = Paths.get(library.libSpec());
-                    if (!useSystemLoadLibrary ||
-                            library.specKind() == Library.SpecKind.NAME ||
-                            (libPath.isAbsolute() && Files.isRegularFile(libPath))) {
-                        builder.addLibrary(library);
-                    } else {
-                        // not an absolute path, but--use-system-load-library was specified
-                        logger.err("l.option.value.absolute.path", lib);
-                    }
-                } catch (IllegalArgumentException ex) {
-                    logger.err("l.option.value.invalid", lib);
-                    return OPTION_ERROR;
-                }
-            }
+        Integer optionError = parseLibraries("l", optionSet, useSystemLoadLibrary, builder);
+        if (optionError != null) return optionError;
+
+        optionError = parseLibraries("f", optionSet, useSystemLoadLibrary, builder);
+        if (optionError != null) return optionError;
+
+
+        if (optionSet.has("--mac-framework-dir")) {
+            optionSet.valuesOf("--mac-framework-dir").forEach(p -> builder.addClangArg("-F" + p));
         }
+
+        inferMacOSFrameworkPath().ifPresent(platformPath -> builder.addClangArg("-F" + platformPath));
 
         String targetPackage = optionSet.has("-t") ? optionSet.valueOf("-t") : "";
         builder.setTargetPackage(targetPackage);
@@ -532,6 +550,31 @@ public final class JextractTool {
                 SUCCESS;
     }
 
+    private Integer parseLibraries(String optionString, OptionSet optionSet, boolean useSystemLoadLibrary, Options.Builder builder) {
+        if (optionSet.has("-" + optionString)) {
+            for (String lib : optionSet.valuesOf("-" + optionString)) {
+                try {
+                    if (optionString.equals("f")) lib = parseFrameworkPath(lib);
+
+                    Library library = Library.parse(lib);
+                    Path libPath = Paths.get(library.libSpec());
+                    if (!useSystemLoadLibrary ||
+                            library.specKind() == Library.SpecKind.NAME ||
+                            (libPath.isAbsolute() && Files.isRegularFile(libPath))) {
+                        builder.addLibrary(library);
+                    } else {
+                        // not an absolute path, but--use-system-load-library was specified
+                        logger.err(optionString + ".option.value.absolute.path", lib);
+                    }
+                } catch (IllegalArgumentException ex) {
+                    logger.err(optionString + ".option.value.invalid", lib);
+                    return OPTION_ERROR;
+                }
+            }
+        }
+        return null;
+    }
+
     /**
      * ToolProvider implementation for jextract tool.
      */
@@ -551,14 +594,24 @@ public final class JextractTool {
     }
 
     private Optional<Path> inferPlatformIncludePath() {
-        String os = System.getProperty("os.name");
-        if (os.equals("Mac OS X")) {
+        return inferMacOsPath("usr", "include");
+    }
+
+    private Optional<Path> inferMacOSFrameworkPath() {
+        return inferMacOsPath("System", "Library", "Frameworks");
+    }
+
+    private static String getMacOsSDKPath() throws IOException {
+        ProcessBuilder pb = new ProcessBuilder().
+                command("/usr/bin/xcrun", "--show-sdk-path");
+        return new String(pb.start().getInputStream().readAllBytes());
+    }
+
+    private Optional<Path> inferMacOsPath(String... paths) {
+        if (isMacOSX) {
             try {
-                ProcessBuilder pb = new ProcessBuilder().
-                    command("/usr/bin/xcrun", "--show-sdk-path");
-                Process proc = pb.start();
-                String str = new String(proc.getInputStream().readAllBytes());
-                Path dir = Paths.get(str.trim(), "usr", "include");
+                String str = getMacOsSDKPath();
+                Path dir = Paths.get(str.trim(), paths);
                 if (Files.isDirectory(dir)) {
                     return Optional.of(dir);
                 }
@@ -568,7 +621,10 @@ public final class JextractTool {
                 }
             }
         }
-
         return Optional.empty();
+    }
+
+    private String parseFrameworkPath(String optionString) {
+        return String.format(":/System/Library/Frameworks/%1$s.framework/%1$s", optionString);
     }
 }
--- a/src/main/java/org/openjdk/jextract/impl/Utils.java
+++ b/src/main/java/org/openjdk/jextract/impl/Utils.java
@@ -27,7 +27,6 @@
 package org.openjdk.jextract.impl;
 
 import org.openjdk.jextract.Declaration;
-import org.openjdk.jextract.Declaration.Constant;
 import org.openjdk.jextract.Type;
 import org.openjdk.jextract.Type.Delegated;
 import org.openjdk.jextract.Type.Delegated.Kind;
@@ -47,6 +46,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Consumer;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * General utility functions
--- a/src/main/resources/org/openjdk/jextract/impl/resources/Messages.properties
+++ b/src/main/resources/org/openjdk/jextract/impl/resources/Messages.properties
@@ -1,21 +1,21 @@
 #
-#  Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+#  Copyright (c) 2020, 2024 Oracle and/or its affiliates. All rights reserved.
 #  DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
-# 
+#
 #  This code is free software; you can redistribute it and/or modify it
 #  under the terms of the GNU General Public License version 2 only, as
 #  published by the Free Software Foundation.
-# 
+#
 #  This code is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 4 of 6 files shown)


src/main/java/org/openjdk/jextract/JextractTool.java
```diff
@@ -532,6 +550,31 @@ private int run(String[] args) {
...
+                    if (optionString.equals("f")) lib = parseFrameworkPath(lib);
+
+                    Library library = Library.parse(lib);
+                    Path libPath = Paths.get(library.libSpec());
+                    if (!useSystemLoadLibrary ||
+                            library.specKind() == Library.SpecKind.NAME ||
+                            (libPath.isAbsolute() && Files.isRegularFile(libPath))) {
+                        builder.addLibrary(library);
+                    } else {
+                        // not an absolute path, but--use-system-load-library was specified
```
Marcono1234, line 566:
```suggestion
                        // not an absolute path, but --use-system-load-library was specified
```

src/main/java/org/openjdk/jextract/JextractTool.java
```diff
@@ -568,7 +621,10 @@ private Optional<Path> inferPlatformIncludePath() {
                 }
             }
         }
-
         return Optional.empty();
     }
+
+    private String parseFrameworkPath(String optionString) {
```
Marcono1234, line 627: Maybe rename to `format...` since this is not parsing anything?
```suggestion
    private String formatFrameworkPath(String optionString) {
```

src/main/java/org/openjdk/jextract/impl/Utils.java
```diff
@@ -47,6 +46,8 @@
 import java.util.List;
 import java.util.Map;
 import java.util.function.Consumer;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
```
Marcono1234, line 50: Unused imports?

src/main/resources/org/openjdk/jextract/impl/resources/Messages.properties
```diff
@@ -83,6 +87,12 @@ Option                             Description
 \                                   is not specified, then unnamed package is used.             \n\
 --version                          print version information and exit                           \n
 
+jextract.usage.mac=\
+Platform dependent options for running jextract                                                 \n\
+--mac-framework-dir <dir>          specify the framework directory                              \n\
+-f <framekwork>                    specify framework library. -f libGL is equivalent to         \n\
```
Marcono1234, line 93:
```suggestion
-f <framework>                    specify framework library. -f libGL is equivalent to         \n\
```

Marcono1234 (commented):
The review comments below are only suggestions, feel free to ignore them, I am not a JDK member. Though hopefully the comments are useful nonetheless.

doc/GUIDE.md
```diff
@@ -988,6 +988,14 @@ A complete list of all the supported command line options is given below:
...
 | `--version`                                                  | print version information and exit |
 
+**macOS platform options (available only when running on macOS):**
+
+| Option                       | Meaning                                                                                                                                                                                                                                 |
+|:-----------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `--mac-framework-dir <dir>`  | specify the framework directory dir include files <br/>defaults to the current Mac OS X SDK dir<br/> This removes the need of having a compile_flags.txt with the required `-framework XYZ` options in the folder where jextract is ran |
```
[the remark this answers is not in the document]
nizarbenalla: The apple gcc manual says

> -Fdir 
Add the framework directory dir to the head of the list of directories to be
searched for header files. These directories are interleaved with those specified
by -I options and are scanned in a left-to-right order.

I tried to follow similar conventions of C compilers

src/main/java/org/openjdk/jextract/JextractTool.java
```diff
@@ -280,6 +290,10 @@ private String singleCharOptionArg(String str) {
             return str.substring(2);
         }
 
+        static boolean checkIfSupported( String option) {
+            return platformOptions.contains(option);
+        }
```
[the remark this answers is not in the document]
nizarbenalla: You're right, I copied the name from the `jpackage` sources but it's not very accurate.

nizarbenalla:
Thanks @Marcono1234 for the comments. They were useful.

src/main/resources/org/openjdk/jextract/impl/resources/Messages.properties
```diff
@@ -83,6 +87,12 @@ Option                             Description
 \                                   is not specified, then unnamed package is used.             \n\
 --version                          print version information and exit                           \n
 
+jextract.usage.mac=\
+Platform dependent options for running jextract                                                 \n\
+--mac-framework-dir <dir>          specify the framework directory                              \n\
+-f <framework>                    specify framework library. -f libGL is equivalent to         \n\
+\                                      -l :/System/Library/Frameworks/libGL.framework/libGL         \n
```
Marcono1234, line 94: Wrong / inconsistent indentation?
```suggestion
-f <framework>                     specify framework library. -f libGL is equivalent to         \n\
\                                   -l :/System/Library/Frameworks/libGL.framework/libGL       \n
```

nizarbenalla: Thanks. Fixed the first line in [9ce83f3](https://github.com/openjdk/jextract/pull/268/commits/9ce83f363e50b6e7f3fe5dae1612b929b38a7bd3), the spacing in the second line is intentional.

doc/GUIDE.md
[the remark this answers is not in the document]
mcimadamore: I believe I agree with @Marcono1234  - it seems like a typo to me.

doc/GUIDE.md
```diff
@@ -988,6 +988,14 @@ A complete list of all the supported command line options is given below:
 | `--include-[function,constant,struct,union,typedef,var]<String>` | Include a symbol of the given name and kind in the generated bindings. When one of these options is specified, any symbol that is not matched by any specified filters is omitted from the generated bindings. |
 | `--version`                                                  | print version information and exit |
 
+**macOS platform options (available only when running on macOS):**
+
+| Option                       | Meaning                                                                                                                                                                                                                                 |
```
mcimadamore, line 993: I'd say let's just add these two options to the list above and maybe put some comment (e.g. MacOS only). If we end up with many platform-specific options we can reconsider of course, but I think there's value in having all the options in the same place.

doc/GUIDE.md
mcimadamore, line 995: Maybe we could call this `macos-framework-dir` ?

mcimadamore: > Maybe we could call this `macos-framework-dir` ?

In fact, I wonder if there could be a more general trick here, where platform-specific options have some kind of prefix - so `macos-framework-dir` and `macos-framework`. What do people think?

src/main/resources/org/openjdk/jextract/impl/resources/Messages.properties
```diff
@@ -83,6 +87,12 @@ Option                             Description
 \                                   is not specified, then unnamed package is used.             \n\
 --version                          print version information and exit                           \n
 
+jextract.usage.mac=\
```
mcimadamore, line 90: Again, not sure it's worth splitting the list of options in two.

src/main/java/org/openjdk/jextract/JextractTool.java
```diff
@@ -532,6 +550,31 @@ private int run(String[] args) {
...
 
+    private Integer parseLibraries(String optionString, OptionSet optionSet, boolean useSystemLoadLibrary, Options.Builder builder) {
+        if (optionSet.has("-" + optionString)) {
+            for (String lib : optionSet.valuesOf("-" + optionString)) {
+                try {
+                    if (optionString.equals("f")) lib = formatFrameworkPath(lib);
+
+                    Library library = Library.parse(lib);
+                    Path libPath = Paths.get(library.libSpec());
+                    if (!useSystemLoadLibrary ||
```
mcimadamore, line 561: Do we accept absolute path for macos frameworks? I thought `-f` only works for _names_ ?

src/main/java/org/openjdk/jextract/JextractTool.java
```diff
@@ -532,6 +550,31 @@ private int run(String[] args) {
                 SUCCESS;
     }
 
+    private Integer parseLibraries(String optionString, OptionSet optionSet, boolean useSystemLoadLibrary, Options.Builder builder) {
```
mcimadamore, line 553: I'd suggest to either use `int` and then `0` for a successful execution, or something like `OptionalInt` if you want to model optionality more explicitly.

src/main/java/org/openjdk/jextract/JextractTool.java
```diff
@@ -568,7 +621,10 @@ private Optional<Path> inferPlatformIncludePath() {
                 }
             }
         }
-
         return Optional.empty();
     }
+
+    private String formatFrameworkPath(String optionString) {
+        return String.format(":/System/Library/Frameworks/%1$s.framework/%1$s", optionString);
```
mcimadamore, line 628: Does this work if a custom framework dir is used? Wouldn't this absolute path point to the wrong place?

mcimadamore (commented):
Great work, I've left some comments for your consideration.

src/main/resources/org/openjdk/jextract/impl/resources/Messages.properties
nizarbenalla: Sure, I did this so that I can show these options only if you're running MacOs

src/main/java/org/openjdk/jextract/JextractTool.java
nizarbenalla: I did this to try to reuse some code.
`-f` is only intended to work with names, otherwise an error should be reported but I will double check.

mcimadamore:
We should make sure that users can specify both `-f` and `-l` and they are _both_ reflected in the generated sources (e.g. in case users would like to generate a single set of sources that work across multiple platforms).

src/main/java/org/openjdk/jextract/JextractTool.java
```diff
@@ -365,6 +384,14 @@ private int run(String[] args) {
         parser.accepts("-t", List.of("--target-package"), "help.t", true);
         parser.accepts("--version", "help.version", false);
 
+        if (isMacOSX) {
+            parser.accepts("--mac-framework-dir", "help.mac.framework", true);
+            parser.accepts("-f", "help.framework.library.path", true);
+        }
+
+        parser.addPlatformOption("--mac-framework-dir");
+        parser.addPlatformOption("-f");
```
JornVernee, line 393: Is `addPlatformOption` just for showing a different error message?

nizarbenalla: Yes, I wanted to have a similar error message to that of `jpackage`.

emrul:
I'm interested in this PR and I gave it a quick test to see if I could import bits of the WebKit Framework. Unless I'm doing something wrong (which is quite possible as its the first time I'm attempting to use jextract...) it appears not to work as expected. 
For example:
```
./jextract --output ./tmp   --target-package com.example.webview.webkit -f Foundation -f WebKit /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/WebKit.framework/Headers/NSAttributedString.h
```

Returns some errors such as:
```
NSObjCRuntime.h:609:19: error: unknown type name 'NSString'
NSObjCRuntime.h:610:53: error: unknown type name 'NSString'
NSObjCRuntime.h:612:19: error: unknown type name 'NSString'
NSObjCRuntime.h:612:50: error: unknown type name 'Protocol'
NSObjCRuntime.h:613:19: error: unknown type name 'Protocol'
```

nizarbenalla:
This PR will only make dealing with MacOs frameworks easier, but if you couldn't generate the Java bindings before this change, you still won't be able to afterward.

jextract will explicitly throw errors for things that are not in the C language (see `TreeMaker::createTree`), I don't think it can process objective-c only constructs out of the box so you will need to make some changes.

nizarbenalla:
Sorry for the delayed update. I've made some changes based on the review comments.
-I've updated the Guide and list of options, they are no longer split into two.
-No longer emit warnings if the mac os options are used on a different platform.
-This feature should work with custom frameworks, I haven't tested it yet but I will try.
-If `--use-system-load-library` option is used without `-l`, a warning is emitted. This can be handled at the parsing level but  checking when adding options to the builder seemed simpler.

doc/GUIDE.md
```diff
@@ -987,13 +987,8 @@ A complete list of all the supported command line options is given below:
...
-| `--mac-framework-dir <dir>`  | specify the framework directory dir include files <br/>defaults to the current Mac OS X SDK dir<br/> This removes the need of having a compile_flags.txt with the required `-framework XYZ` options in the folder where jextract is ran |
-| `-f, framework <framework>` | specify the name of the library, path will be expanded to that of the framework folder                                                                                                                                                  |
+| `--macos-framework <dir>` (macOs only)           | specify the framework directory include files <br/>defaults to the current Mac OS X SDK dir<br/> This removes the need of having a compile_flags.txt with the required `-framework XYZ` options in the folder where jextract is ran |
+| `-f <framework>` (macOs only)                    | specify the name of the library, path will be expanded to that of the framework folder                                                                                                                                                 |
```
JornVernee, line 991: I don't think we should mention the workaround using `compile_flags.txt` here. This won't make sense to users who are not familiar with it, and after this change, there's no need for the workaround.

src/main/java/org/openjdk/jextract/JextractTool.java
```diff
@@ -383,14 +360,8 @@ private int run(String[] args) {
...
-
-        if (isMacOSX) {
-            parser.accepts("--mac-framework-dir", "help.mac.framework", true);
-            parser.accepts("-f", "help.framework.library.path", true);
-        }
-
-        parser.addPlatformOption("--mac-framework-dir");
-        parser.addPlatformOption("-f");
+        parser.accepts("--macos-framework", "help.mac.framework", true);
+        parser.accepts("-f", "help.framework.library.path", true);
```
JornVernee, line 364: In clang these are named `-F` and `-framework` [1] (there's no `-f`). I think we should just match that. We did the same for `-D`, `-I` and `-l` to an extent.

[1]: https://clang.llvm.org/docs/ClangCommandLineReference.html

src/main/java/org/openjdk/jextract/JextractTool.java
```diff
@@ -625,6 +601,15 @@ private Optional<Path> inferMacOsPath(String... paths) {
     }
 
     private String formatFrameworkPath(String optionString) {
-        return String.format(":/System/Library/Frameworks/%1$s.framework/%1$s", optionString);
+        String publicPath = String.format("/System/Library/Frameworks/%1$s.framework/", optionString);
+        String privatePath = String.format("/System/Library/PrivateFrameworks/%1$s.framework/", optionString);
```
JornVernee, line 605: I think Maurizios comment still applies. AFAIU we need to pass the exact framework path to `dlopen`, so we need to manually search for the frameworks in this system location, and also in the locations specified through `-F`, and then load it from one of those locations.

src/main/java/org/openjdk/jextract/JextractTool.java
mcimadamore: I mean -- if one has a prefix `macos-` the other should too -- or none should have it.

mcimadamore: Also, `-f` usually is used in conjunction with other options e.g. `-fmove-loop-invariants`. Which is probably the reason behind the capital letter `-F`.

mcimadamore: I'm fine with that

src/main/java/org/openjdk/jextract/JextractTool.java
```diff
@@ -360,8 +362,8 @@ private int run(String[] args) {
         parser.accepts("--output", "help.output", true);
         parser.accepts("-t", List.of("--target-package"), "help.t", true);
         parser.accepts("--version", "help.version", false);
-        parser.accepts("--macos-framework", "help.mac.framework", true);
-        parser.accepts("-f", "help.framework.library.path", true);
+        parser.accepts("-framework", "help.mac.framework", true);
+        parser.accepts("-F", "help.framework.library.path", true);
```
JornVernee, line 366: I think these should be the other way around. `-F` specifies the search path, and `-framework` accepts the name of the framework you want to link against (e.g. `-framework OpenGL`).

src/main/java/org/openjdk/jextract/JextractTool.java
```diff
@@ -601,6 +612,13 @@ private Optional<Path> inferMacOsPath(String... paths) {
...
     private String formatFrameworkPath(String optionString) {
+        for (String dir : frameworkDirs) {
+            Path path = Path.of(dir, optionString + ".framework");
+            if (Files.exists(path)) {
+                return String.format(":%s%s", path, optionString);
+            }
+        }
+
         String publicPath = String.format("/System/Library/Frameworks/%1$s.framework/", optionString);
         String privatePath = String.format("/System/Library/PrivateFrameworks/%1$s.framework/", optionString);
```
JornVernee, line 623: Also, should this default just use the result of `inferMacOSFrameworkPath`?

src/main/java/org/openjdk/jextract/JextractTool.java
```diff
@@ -389,7 +391,10 @@ private int run(String[] args) {
         Path compileFlagsTxt = Paths.get(".", "compile_flags.txt");
         if (Files.exists(compileFlagsTxt)) {
             try {
-                Files.lines(compileFlagsTxt).forEach(opt -> builder.addClangArg(opt));
+                Files.lines(compileFlagsTxt).forEach(opt -> {
+                    builder.addClangArg(opt);
+                    frameworkDirs.add(opt.substring(2).trim());
```
JornVernee, line 396: Compile flags can also contain other flags, so we have to explicitly look for `-F` options here for this to work.

But, I'd also be okay just ignoring the `-F` flags in `compile_flags`. We can tell people to use the builtin `-F`.

src/main/java/org/openjdk/jextract/JextractTool.java
```diff
@@ -444,15 +449,22 @@ private int run(String[] args) {
...
+        optionError = parseLibraries("F", optionSet, useSystemLoadLibrary, builder);
         if (optionError != 0) return optionError;
 
 
-        if (optionSet.has("--macos-framework")) {
-            optionSet.valuesOf("--macos-framework").forEach(p -> builder.addClangArg("-F" + p));
+        if (optionSet.has("-framework")) {
+            optionSet.valuesOf("-framework").forEach(
+                    p -> {
+                        builder.addClangArg("-F" + p);
```
JornVernee, line 459: See the mismatch here. Both should be `-F`.

src/main/java/org/openjdk/jextract/JextractTool.java
nizarbenalla: Sure, note that the result of `inferMacOSFrameworkPath ` is checked before as it was added to `frameworkDirs`.
The path would still be incorrect but it's better than a malformed path.

src/main/java/org/openjdk/jextract/JextractTool.java
nizarbenalla: I chose to ignore the `-F` flags in `compile_flags`.
The paths passed using the builtin `-F` will be checked first before public/private paths and the MacOs SDK, that way there "good" defaults and it's simple enough for users to override them

src/main/java/org/openjdk/jextract/JextractTool.java
```diff
@@ -601,15 +606,20 @@ private Optional<Path> inferMacOsPath(String... paths) {
     }
 
     private String formatFrameworkPath(String optionString) {
+        for (String dir : frameworkDirs) {
```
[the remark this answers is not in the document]
nizarbenalla: I've made a small update to achieve this. Also moved the code the handle the `-F` options to be higher as we need to add those paths to the list before we can resolve the framework paths. (I missed this before)

I've renamed some variables/methods to be more accurate, as the previous name was a bit misleading.

src/main/java/org/openjdk/jextract/JextractTool.java
```diff
@@ -533,8 +536,12 @@ private int parseLibraries(String optionString, OptionSet optionSet, boolean use
         if (optionSet.has("-" + optionString)) {
             for (String lib : optionSet.valuesOf("-" + optionString)) {
                 try {
-                    if (optionString.equals("framework")) {
-                        lib = formatFrameworkPath(lib);
+                    lib = optionString.equals("framework") ?
```
mcimadamore, line 539: Is the argument of `framework` always a relative path? Or does macos also allow absolute paths to be used? In that case, should be check for something that starts with `/` ?

mcimadamore: gaaaah - got confused -- `-framework` is a framework name. In that case, the code is good (and better than before, thanks!). But the exception seems misleading. It should say something like `Cannot resolve/find framework <name>`, rather than speaking about paths?

nizarbenalla: Updated the error message, the user should see `error: cannot find framework for -framework option: NAME`

src/main/resources/org/openjdk/jextract/impl/resources/Messages.properties
```diff
@@ -26,7 +26,7 @@ argfile.read.error=reading @argfile failed: {0}
 cannot.read.header.file=cannot read header file: {0}
 not.a.file=not a file: {0}
 l.option.value.invalid=invalid library specifier for -l option: {0}
-framework.option.value.invalid=invalid library specifier for -framework option: {0}
+framework.option.value.invalid=cannot find framework for -framework option: {0}
```
mcimadamore, line 29: I suggest to just say `Cannot find framework {0}`. Adding `for framework option` doesn't seem too useful?

nizarbenalla: Sure, I admit it's redundant information

mcimadamore (approved):
Looks good - thanks!

src/main/java/org/openjdk/jextract/JextractTool.java
```diff
@@ -439,30 +442,27 @@ private int run(String[] args) {
...
-                    logger.err("l.option.value.invalid", lib);
-                    return OPTION_ERROR;
-                }
-            }
+        if (optionSet.has("-F")) {
+            List<String> paths = optionSet.valuesOf("-F");
+
+            paths.forEach(p -> builder.addClangArg("-F" + p));
+            frameworkPaths.addAll(0, paths);
         }
```
JornVernee, line 456: My only concern left is what now happens on non-mac platforms when `-F` or `-framework` is used. I suppose clang might error out on the flag in the case of `-F`, not sure what happens for `-framework`.

I think these options should perhaps be disabled on platforms besides mac, since they are only intended to be used there.

src/main/java/org/openjdk/jextract/JextractTool.java
```diff
@@ -364,6 +367,8 @@ private int run(String[] args) {
         parser.accepts("--output", "help.output", true);
         parser.accepts("-t", List.of("--target-package"), "help.t", true);
         parser.accepts("--version", "help.version", false);
+        parser.accepts("-F", "help.mac.framework", true);
+        parser.accepts("-framework", "help.framework.library.path", true);
```
JornVernee, line 371: e.g. something simple like:
```suggestion
        if (isMacOSX) {
            parser.accepts("-F", "help.mac.framework", true);
            parser.accepts("-framework", "help.framework.library.path", true);
        }
```

mcimadamore:
Also forgot about this:
https://github.com/openjdk/jextract/blob/master/samples/opengl/compile_flags.txt

The OpenGL example has a compile_flags to set the framework option. I think that should be dropped and the new jextract option should be mentioned in that example

src/main/java/org/openjdk/jextract/JextractTool.java
nizarbenalla: @mcimadamore What do you think about emitting errors on non-mac platforms?

mcimadamore:
> Also forgot about this: https://github.com/openjdk/jextract/blob/master/samples/opengl/compile_flags.txt
> 
> The OpenGL example has a compile_flags to set the framework option. I think that should be dropped and the new jextract option should be mentioned in that example

Apologies -- I've missed the changes to the example -- this looks good already

src/main/java/org/openjdk/jextract/JextractTool.java
mcimadamore: I think we should either:
* not recognize the options when not on macos
* recognize them, but turn them into no-op (e.g. do nothing if not on macos)
If we recognize them AND try to act on them (e.g. by setting `-F` on the underlying `clang` process, I could see issue with that approach, as @JornVernee  mentions)

JornVernee (approved):
Latest version looks good

nizarbenalla:
Thanks for all the reviews.

/integrate

PR #269, opened 2024-12-30
7903923: Derive C_* layouts directly from the Linker

nizarbenalla:
Depending on the platform, different `C_*`  correspond to different `JAVA_*` types. We can get those values directly from the linker.

This patch can allows us to have a common class for different platforms.

Please review, and thanks in advance!

---------
### Progress

### Issue
 * [CODETOOLS-7903923](https://bugs.openjdk.org/browse/CODETOOLS-7903923): Derive C_* layouts directly from the Linker (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/269/head:pull/269` \
`$ git checkout pull/269`

`$ git checkout pull/269` \
`$ git pull https://git.openjdk.org/jextract.git pull/269/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 269`

`$ git pr show -t 269`

</details>

<a href="https://git.openjdk.org/jextract/pull/269.diff">https://git.openjdk.org/jextract/pull/269.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/269#issuecomment-2565682345)
</details>


src/main/java/org/openjdk/jextract/impl/ToplevelBuilder.java
```diff
@@ -66,21 +66,21 @@ private static HeaderFileBuilder createFirstHeader(SourceFileBuilder sfb, List<O
...
+            public static final ValueLayout.OfLong C_LONG_LONG = (ValueLayout.OfLong) Linker.nativeLinker().canonicalLayouts().get("long");
+            public static final ValueLayout.OfFloat C_FLOAT = (ValueLayout.OfFloat) Linker.nativeLinker().canonicalLayouts().get("float");
+            public static final ValueLayout.OfDouble C_DOUBLE = (ValueLayout.OfDouble) Linker.nativeLinker().canonicalLayouts().get("double");
+            public static final AddressLayout C_POINTER =((AddressLayout) Linker.nativeLinker().canonicalLayouts().get("void*"))
                     .withTargetLayout(MemoryLayout.sequenceLayout(java.lang.Long.MAX_VALUE, JAVA_BYTE));
             """);
         if (TypeImpl.IS_WINDOWS) {
-            first.appendIndentedLines("public static final ValueLayout.OfInt C_LONG = ValueLayout.JAVA_INT;");
-            first.appendIndentedLines("public static final ValueLayout.OfDouble C_LONG_DOUBLE = ValueLayout.JAVA_DOUBLE;");
+            first.appendIndentedLines("public static final ValueLayout.OfInt C_LONG = (ValueLayout.OfInt) Linker.nativeLinker().canonicalLayouts().get(\"int\");");
```
mcimadamore, line 80: The canonical layout for this has name `long`. But the result is a `ValueLayout.OfInt` (because FFM uses `int` to model values of that type on Windows).

src/main/java/org/openjdk/jextract/impl/ToplevelBuilder.java
```diff
@@ -66,21 +66,21 @@ private static HeaderFileBuilder createFirstHeader(SourceFileBuilder sfb, List<O
...
+            first.appendIndentedLines("public static final ValueLayout.OfDouble C_LONG_DOUBLE = (ValueLayout.OfDouble) Linker.nativeLinker().canonicalLayouts().get(\"double\");");
```
mcimadamore, line 81: Here, there's no canonical layout for `long double`. (but what you did retains compatibility with what we had). But I wonder if FFM should add one? @JornVernee , @minborg  ?

JornVernee: I would be okay with adding `long double` as a canonical layout, just on Windows, if that's what you mean.

src/main/java/org/openjdk/jextract/impl/ToplevelBuilder.java
nizarbenalla: Thanks for explaining, will fix it.

nizarbenalla: Fixed in ed192bbf9dbc39e1871722b0470a73f14a8a0c58

nizarbenalla:
I'm seeing multiple failures on windows (and only on windows), will need to investigate.

nizarbenalla:
I now use the correct layout for `C_LONG_LONG`. Passes tier 1 on Mac and Windows.

mcimadamore:
What's the story with primitive typedef? Do they work? I believe they will go back to stuff like JAVA_INT and such, instead of picking the layout constant we have already emitted. E.g. 

`typedef int myInt`

mcimadamore:
> What's the story with primitive typedef? Do they work? I believe they will go back to stuff like JAVA_INT and such, instead of picking the layout constant we have already emitted. E.g.
> 
> `typedef int myInt`

I think to address this, the method ClassSourceBuilder::primitiveLayoutString might come in useful, as we use that method to generate references to primitive layouts e.g. inside struct layout declarations.

nizarbenalla:
> What's the story with primitive typedef? Do they work? I believe they will go back to stuff like JAVA_INT and such, instead of picking the layout constant we have already emitted. E.g.
>
> typedef int myInt

I do not think so. Instead `C_INT` derived from the canonical layout is used everywhere

```
cat test.h 
typedef long myLong;
typedef int myInt;
```
`jextract` generates the following code 

```
public static final OfLong myLong = test_h.C_LONG;
public static final OfInt myInt = test_h.C_INT;
```
I generated a couple of samples (sqlite3 and opengl) and do not see `JAVA_INT` anywhere, the only place where this string is found in the repo is the generated libclang folder.

nizarbenalla:
Made a small change, I wanted to drop the last usage of `JAVA_*` that remained.

mcimadamore (approved):
Looks good!

nizarbenalla:
Thanks all for the reviews on this.

/integrate

PR #278, opened 2025-02-03
7903933: Move sharable items from different generations to a common file

nizarbenalla:
Please review this patch to move the `C_*` layouts and the static utility methods into separate classes: `LayoutUtils.java` and `FFMUtils.java`, respectively.

- The names could later be personalized through a JSON configuration.
- We can use static imports if the `-t` option is no used and the files are generated into the default package, in that case we use the classname to call the static methods or use the `C_*` constants.

Some tests had to be modified slightly, either by adding new static imports or replacing classnames.

---------
### Progress

### Issue
 * [CODETOOLS-7903933](https://bugs.openjdk.org/browse/CODETOOLS-7903933): Move sharable items from different generations to a common file (**Enhancement** - P4)

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - **Reviewer**) ⚠️ Review applies to [e3b9ae2a](https://git.openjdk.org/jextract/pull/278/files/e3b9ae2a21517e008a06e3e55d42a5957366d4ad)
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/278/head:pull/278` \
`$ git checkout pull/278`

`$ git checkout pull/278` \
`$ git pull https://git.openjdk.org/jextract.git pull/278/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 278`

`$ git pr show -t 278`

</details>

<a href="https://git.openjdk.org/jextract/pull/278.diff">https://git.openjdk.org/jextract/pull/278.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/278#issuecomment-2631818307)
</details>


nizarbenalla:
I wasn't happy with the previous approach/implementation and started over from scratch, also did some small cleanup (the newly added option for framworks should use `--` rather than `-`).

Shared items from multiple generation will only be moved to a class if the user specifies the `--sharable-items` option, with the class name of his choosing and the main header file extends that class. This will be helpful for project that need to run jextract on multiple headers, especially if those headers are large.

mcimadamore:
Adding an option to move shared items in a separate class feels like taking a shortcut. Let's look what's inside this bag of shared items (it used to be much bigger, so good news there):

1. library arena
2. trace downcall support
3. alignment method
4. findOrThrow
5. upcallHandle
6. primitive layout constants

I believe (1) should probably not belong in a separate class from the headers. The way I see it is that it should belong where the library lookups belongs to. Even if you run multiple extractions it is less clear to me as to whether there should be a single lifetime for all the different libraries or not (and, eventually, we want aspects such as this to be customizable via subclassing).

(2) seems a very useful debug option, and I think we should think about making it more official -- for instance with a linker option.

(3) is a general purpose method that would make sense to feature in the main MemoryLayout API.

(4) seems to be superseded by a similar method that was added in Java 23:
https://docs.oracle.com/en/java/javase/23/docs//api/java.base/java/lang/foreign/SymbolLookup.html#findOrThrow(java.lang.String)

(5) seems like it could be inlined in the functional interface classes (it's just an helper to allow clients to create an upcall stub w/o the need to do a MethodHandle lookup -- which needs a try/catch).

Then there's (6). The first reaction I got was: well, whether primitive layouts should be emitted or not seems like another filtering decision (e.g. let's add some options to filter these out). Except this would not work -- it's not just about filtering -- it's also about telling every other file where to find the layouts for such primitive types. If they are defined somewhere else, then jextract need to know _where_ to find them (e.g. if they are referenced by some other layout jextract is building). Which seems a similar problem as the one this PR is trying to solve anyway.

Stepping back -- I think if we look back, jetxract used to generate a RuntimeHelper class, and then everything else. While we moved away from that generation scheme (now we just emit header classes), the shared functionalities are still there. Perhaps a move in a good direction would be to:

* put all the shared functionalities in the root of the header hierarchy -- and _don't put anything else in there_
* this means that there will always be at least two header classes generated `foo_f` and `foo_h$0`, where `foo_h extends foo_h$0` and all the shared symbols are in `foo_h$0`. 
* add an option to override the name of that root header class

This is similar, in spirit, to what you have here, but with the advantage that there's only one generation scheme, not two -- e.g. the superclass with the shared symbol is _always_ there -- only sometimes it can have a different name (because the user said so). Whether we want that default superclass name to be `foo_h$0`, or maybe something more explicit like `foo_h_shared`, I'm open to suggestions.

But I do think that we should try and make that shared class as small as possible -- most of the functionality there seems like it could belong in the main FFM API, and it would provide value even for clients not running on top of jextract.

mcimadamore:
> Then there's (6). The first reaction I got was: well, whether primitive layouts should be emitted or not seems like another filtering decision (e.g. let's add some options to filter these out). Except this would not work -- it's not just about filtering -- it's also about telling every other file where to find the layouts for such primitive types. If they are defined somewhere else, then jextract need to know _where_ to find them (e.g. if they are referenced by some other layout jextract is building). Which seems a similar problem as the one this PR is trying to solve anyway.

Note/history: while tempting, we can't really put 6 inside the main FFM API. We have thought about this for a long time -- the issue is that the types of the primitive layouts is not guaranteed to be stable. E.g. `C_LONG` might be either a `ValueLayout.OfLong` or a `ValueLayout.OfInt`, depending on the platforms. Other layouts, such as `C_LONG_DOUBLE` might only be available on some platforms and not others. This is why, long ago, we decided that the main FFM API should not concern with providing layout constants for C types -- as the set of such constant is not stable. Instead, such primitive C layouts can be "discovered" using the `Linker::canonicalLayouts` API.

mcimadamore:
> Then there's (6). The first reaction I got was: well, whether primitive layouts should be emitted or not seems like another filtering decision (e.g. let's add some options to filter these out). Except this would not work -- it's not just about filtering -- it's also about telling every other file where to find the layouts for such primitive types. If they are defined somewhere else, then jextract need to know _where_ to find them (e.g. if they are referenced by some other layout jextract is building). Which seems a similar problem as the one this PR is trying to solve anyway.

Also on this topic: for now we're mostly concerned about different extractions not repeating the code for primitive layouts and helper functions. This feels more like a "tip of the iceberg" kind of situation. For instance, you might have two libraries A and B, which both include the header of some third library C. Maybe you want to extract C separately, and then extract A and B so that they somehow magically point at the extracted bindings for C. Now sharing would be not just about primitive types, but about functions, structs and much more.

At the same time, going down this path can be very complex: A and B might pull in slightly different versions of C, or use some `#define` macro directives which would alter the shape of the generated bindings in C. In which case reusing the same bindings for C would be more difficult.

So, hidden somewhere in here there's a theme of: how do we move jextract to go from a per-extraction set of bindings to a multi-extraction friendly model. And going down this path will likely, I think, result in opening a big and complicated can of worms. I'm not saying we'll never get there -- but there's a limit with what we can express with simple command line options.

JornVernee:
> this means that there will always be at least two header classes generated foo_f and foo_h$0, where foo_h extends foo_h$0 and all the shared symbols are in foo_h$0.

Should we just give the base header a common name then, so that if you generate multiple times, you get sharing automatically? Maybe it could be named something like `Builtins`.

mcimadamore:
> > this means that there will always be at least two header classes generated foo_f and foo_h$0, where foo_h extends foo_h$0 and all the shared symbols are in foo_h$0.
> 
> Should we just give the base header a common name then, so that if you generate multiple times, you get sharing automatically? Maybe it could be named something like `Builtins`.

Yes, see

> This is similar, in spirit, to what you have here, but with the advantage that there's only one generation scheme, not two -- e.g. the superclass with the shared symbol is always there -- only sometimes it can have a different name (because the user said so). Whether we want that default superclass name to be foo_h$0, or maybe something more explicit like foo_h_shared, I'm open to suggestions.

nizarbenalla:
> (4) findOrThrow
> (4) seems to be superseded by a similar method that was added in Java 23:
https://docs.oracle.com/en/java/javase/23/docs//api/java.base/java/lang/foreign/SymbolLookup.html#findOrThrow(java.lang.String)

I meant to remove this when targeting jdk 23, it seems I left it by mistake. I will remove it to reduce the number of shared items we're dealing with.

nizarbenalla:
I've moved the shared symbols to a different class `foo_h$shared` that all headers extend from.

I've also added a new command line option `--shared-symbols` to allow users to specify the name of the class if they want to.

mcimadamore:
The option for changing header class name is called `--header-class-name <name>`. So I'd suggest something like `--symbols-class-name`.

src/main/java/org/openjdk/jextract/impl/ToplevelBuilder.java
```diff
@@ -51,39 +52,33 @@ class ToplevelBuilder implements OutputFactory.Builder {
...
+        otherBuilders.add(sfb);
+    }
+
     private static HeaderFileBuilder createFirstHeader(SourceFileBuilder sfb, List<Options.Library> libs, boolean useSystemLoadLibrary) {
-        HeaderFileBuilder first = new HeaderFileBuilder(sfb, String.format("%1$s#{SUFFIX}",sfb.className()), null, sfb.className());
+        HeaderFileBuilder first = new HeaderFileBuilder(sfb, String.format("%1$s#{SUFFIX}",sfb.className()), SHARED, sfb.className());
         first.appendBlankLine();
         first.classBegin();
         first.emitDefaultConstructor();
         first.emitRuntimeHelperMethods();
```
mcimadamore, line 80: Shouldn't the runtime helper methods end up in the shared class?

src/main/java/org/openjdk/jextract/impl/ToplevelBuilder.java
```diff
@@ -51,39 +52,33 @@ class ToplevelBuilder implements OutputFactory.Builder {
...
+        createSharedClass();
         SourceFileBuilder sfb = SourceFileBuilder.newSourceFile(packageName, headerClassName);
         headerBuilders.add(sfb);
         lastHeader = createFirstHeader(sfb, libs, useSystemLoadLibrary);
     }
 
+    private void createSharedClass() {
+        SourceFileBuilder sfb = SourceFileBuilder.newSourceFile(packageName(), SHARED);
+        HeaderFileBuilder sharedBuilder = new HeaderFileBuilder(sfb, SHARED, null, SHARED);
+        sharedBuilder.appendBlankLine();
```
mcimadamore, line 67: There seems to be a lot of overlapping between the code here and the one in `nextHeader` -- especially now that the first header is not really meant to have a lot of special things (the only exception being the symbol lookup logic).

src/main/java/org/openjdk/jextract/impl/ToplevelBuilder.java
nizarbenalla: Fixed in https://github.com/openjdk/jextract/pull/278/commits/5c8802a6abc2fbb65a1bf341be2e7e11c5936c74, I thought only the shared symbols belonged in the separate class.

I did not move the library arena to the separate class and instead kept it close to the symbol lookup.

src/main/java/org/openjdk/jextract/impl/ToplevelBuilder.java
nizarbenalla: Fixed, I consolidated the shared code in some methods. I ran into issues with the java files having the wrong name so I had to change the logic a little.

mcimadamore (approved):
Very nice cleanup! Looks good

nizarbenalla:
I pushed a small test-only update, I was seeing some failures in the CI (that I didn't see locally). This fixes it,

mcimadamore (approved):
Good catch!

nizarbenalla:
Thanks for all the round of reviews.

/integrate

src/main/java/org/openjdk/jextract/impl/HeaderFileBuilder.java
```diff
@@ -351,10 +351,14 @@ void emitFirstHeaderPreamble(List<Options.Library> libraries, boolean useSystemL
                 .collect(Collectors.joining(String.format("\n%1$s", indentString(2)), "static final SymbolLookup SYMBOL_LOOKUP = ", ";")));
     }
 
+    void emitLibaryArena(){appendIndentedLines("""
+
+            static final Arena LIBRARY_ARENA = Arena.ofAuto();""");
+    }
```
JornVernee, line 357:
```suggestion
    void emitLibaryArena(){
        appendIndentedLines("""
            static final Arena LIBRARY_ARENA = Arena.ofAuto();""");
    }
```

src/main/java/org/openjdk/jextract/impl/IncludeHelper.java
```diff
@@ -84,6 +84,11 @@ static IncludeKind fromScoped(Declaration.Scoped scoped) {
     private final EnumMap<IncludeKind, Set<String>> includesSymbolNamesByKind = new EnumMap<>(IncludeKind.class);
     private final Set<Declaration> usedDeclarations = new HashSet<>();
     public String dumpIncludesFile;
+    public String sharedSymbolsFile;
+
+    public String getSharedSymbolsFile() {
+        return sharedSymbolsFile;
+    }
```
JornVernee, line 91: Uhm, I think this should just be in `Options`, not in `IncludeHelper`, since it's not related to the `--include` options.

src/main/java/org/openjdk/jextract/impl/ToplevelBuilder.java
```diff
@@ -43,48 +43,76 @@ class ToplevelBuilder implements OutputFactory.Builder {
     private static final int DECLS_PER_HEADER_CLASS = Integer.getInteger("jextract.decls.per.header", 1000);
     public static final String PREV_SUFFIX = "#{PREV_SUFFIX}";
     private static final String SUFFIX = "#{SUFFIX}";
+    private static String SHARED;
```
JornVernee, line 46: I think this was supposed to be `final`?
```suggestion
    private final String shared;
```

doc/GUIDE.md
```diff
@@ -975,20 +975,21 @@ glutSolidTeapot(0.5)
...
+| `--dump-includes <String>`                                       | dump included symbols into specified file (see below)                                                                                                                                                                                                                                 |
+| `--include-[function,constant,struct,union,typedef,var]<String>` | Include a symbol of the given name and kind in the generated bindings. When one of these options is specified, any symbol that is not matched by any specified filters is omitted from the generated bindings.                                                                        |
+| `--symbols-class-name <name>`                                          | override the name of the root header class                                                                                                                                                                                                                                            |
```
JornVernee, line 989:
```suggestion
| `--symbols-class-name <name>`                                    | override the name of the root header class                                                                                                                                                                                                                                            |
```

src/main/resources/org/openjdk/jextract/impl/resources/Messages.properties
```diff
@@ -84,6 +85,7 @@ Option                             Description
 \                                   option is not specified, then current directory is used.    \n\
 -t, --target-package <package>     target package name for the generated classes. If this option\n\
 \                                   is not specified, then unnamed package is used.             \n\
+--symbols-class-name <name>            override the name of the root header class                   \n\
```
JornVernee, line 88:
```suggestion
--symbols-class-name <name>        override the name of the root header class                   \n\
```

src/main/java/org/openjdk/jextract/impl/FunctionalInterfaceBuilder.java
```diff
@@ -42,7 +42,7 @@ final class FunctionalInterfaceBuilder extends ClassSourceBuilder {
 
     private FunctionalInterfaceBuilder(SourceFileBuilder builder, String className, ClassSourceBuilder enclosing,
                                        String runtimeHelperName, Type.Function funcType, boolean isNested) {
-        super(builder, isNested ? "public static" : "public", Kind.CLASS, className, null, enclosing, runtimeHelperName);
+        super(builder, isNested ? "public final static" : "public final", Kind.CLASS, className, null, enclosing, runtimeHelperName);
```
JornVernee, line 45: Changes in this file seem unrelated? Leftover from other work?

test/testng/org/openjdk/jextract/test/toolprovider/TestClassGeneration.java
```diff
@@ -192,7 +192,7 @@ public void testStructMember(String structName, MemoryLayout memberLayout, Class
     @Test(dataProvider = "functionalInterfaces")
     public void testFunctionalInterface(String name, MethodType type) {
         Class<?> fpClass = loader.loadClass("com.acme." + name);
-        checkDefaultConstructor(fpClass);
+        checkPrivateConstructor(fpClass);
```
JornVernee, line 195: Same here, these changes seem unrelated.

src/main/java/org/openjdk/jextract/impl/FunctionalInterfaceBuilder.java
nizarbenalla: This change was to make the utility classes final, it's somewhat unrelated to this patch but it didn't seem to be worth a separate PR.

test/testng/org/openjdk/jextract/test/toolprovider/TestClassGeneration.java
nizarbenalla: This is also related to making the utility classes final with a private constructor.

src/main/java/org/openjdk/jextract/impl/IncludeHelper.java
```diff
@@ -84,10 +84,10 @@ static IncludeKind fromScoped(Declaration.Scoped scoped) {
     private final EnumMap<IncludeKind, Set<String>> includesSymbolNamesByKind = new EnumMap<>(IncludeKind.class);
     private final Set<Declaration> usedDeclarations = new HashSet<>();
     public String dumpIncludesFile;
-    public String sharedSymbolsFile;
+    public String sharedClassName;
 
-    public String getSharedSymbolsFile() {
-        return sharedSymbolsFile;
+    public String getSharedClassName() {
+        return sharedClassName;
```
JornVernee, line 90: These can now be removed I think?

JornVernee (approved):

src/main/java/org/openjdk/jextract/impl/IncludeHelper.java
nizarbenalla: I missed this, thanks for catching it.

JornVernee (approved):

JornVernee:
Thanks for the updates. Looks good now!

nizarbenalla:
Thanks! Glad this work can be integrated into jextract.

/integrate

PR #279, opened 2025-03-28
7903984: Inline upcallHandle method to reduce shared code across multiple generations

nizarbenalla:
Please review this patch to inline the upcallHandle method, reducing the amount of shared items across multiple jextract generations.

Edit: all tests pass in CI, GitHub failure is unrelated 

TIA

---------
### Progress

### Issue
 * [CODETOOLS-7903984](https://bugs.openjdk.org/browse/CODETOOLS-7903984): Inline upcallHandle method to reduce shared code across multiple generations (**Enhancement** - P4)

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - **Reviewer**) ⚠️ Review applies to [3d387faf](https://git.openjdk.org/jextract/pull/279/files/3d387fafb85cfa5848a04fb3f778ebad46037847)
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to [273f5f22](https://git.openjdk.org/jextract/pull/279/files/273f5f22b4accfe74d806cb9ad430d4c71ae89c2)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/279/head:pull/279` \
`$ git checkout pull/279`

`$ git checkout pull/279` \
`$ git pull https://git.openjdk.org/jextract.git pull/279/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 279`

`$ git pr show -t 279`

</details>

<a href="https://git.openjdk.org/jextract/pull/279.diff">https://git.openjdk.org/jextract/pull/279.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/279#issuecomment-2761506259)
</details>

### DIFF
--- a/src/main/java/org/openjdk/jextract/impl/FunctionalInterfaceBuilder.java
+++ b/src/main/java/org/openjdk/jextract/impl/FunctionalInterfaceBuilder.java
@@ -42,7 +42,7 @@ final class FunctionalInterfaceBuilder extends ClassSourceBuilder {
 
     private FunctionalInterfaceBuilder(SourceFileBuilder builder, String className, ClassSourceBuilder enclosing,
                                        String runtimeHelperName, Type.Function funcType, boolean isNested) {
-        super(builder, isNested ? "public static" : "public", Kind.CLASS, className, null, enclosing, runtimeHelperName);
+        super(builder, isNested ? "public final static" : "public final", Kind.CLASS, className, null, enclosing, runtimeHelperName);
         this.parameterNames = funcType.parameterNames().map(NameMangler::javaSafeIdentifiers);
         this.funcType = funcType;
         this.methodType = Utils.methodTypeFor(funcType);
@@ -55,7 +55,7 @@ final class FunctionalInterfaceBuilder extends ClassSourceBuilder {
         fib.appendBlankLine();
         fib.emitDocComment(parentDecl);
         fib.classBegin();
-        fib.emitDefaultConstructor();
+        fib.emitPrivateConstructor();
         String fiName = fib.emitFunctionalInterface();
         fib.emitDescriptorDecl();
         fib.emitFunctionalFactory(fiName);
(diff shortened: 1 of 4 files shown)


mcimadamore (approved):
Looks good. Let's also hear from @JornVernee

(when looking at some generated code I also noted that the functional interface code already has a try/catch block in the `invoke` method).

I guess the unfortunate issue here is that the use is from a static initializer - which means we either need to put the logic in a method (as you have done) or use a static initializer. So we pay a bit more than just for the single try/catch block.

One thing I noticed is that `ConstantBootstraps` has bootstraps to make field var handles (static and non-static) as well as array element var handles. I wonder if we should double down on those and add a way to create direct method handles too... these methods don't throw checked exception, so they are more amenable to generated code like this.

nizarbenalla:
Pushed a new commit to resolve merge conflict.

src/main/java/org/openjdk/jextract/impl/FunctionalInterfaceBuilder.java
```diff
@@ -83,7 +83,15 @@ public interface %1$s {
...
-            private static final MethodHandle UP$MH = %1$s.upcallHandle(%2$s.%3$s.class, "apply", $DESC);
+            static MethodHandle upcallHandle(Class<?> fi, String name, FunctionDescriptor fdesc) {
+                try {
+                    return MethodHandles.lookup().findVirtual(fi, name, fdesc.toMethodType());
+                } catch (ReflectiveOperationException ex) {
+                    throw new AssertionError(ex);
+                }
+            }
+
+            private static final MethodHandle UP$MH = upcallHandle(%2$s.%3$s.class, "apply", $DESC);
```
JornVernee, line 94: Maybe you can inline the arguments into the implementation, since they seem to be constant?

```suggestion
            static MethodHandle upcallHandle() {
                try {
                    return MethodHandles.lookup().findVirtual(%2$s.%3$s.class, "apply", $DESC.toMethodType());
                } catch (ReflectiveOperationException ex) {
                    throw new AssertionError(ex);
                }
            }

            private static final MethodHandle UP$MH = upcallHandle();
```

JornVernee (approved):
Looks good.

To repeat some of the offline discussion: I don't think `ConstantBootstraps` is the right class to add a helper, since that class contains runtime support methods for implementing dynamic constants. However, if you wanted a MethodHandle constant, you could just use a `CONSTANT_MethodHandle_info` instead. I think if we added a more general API, there would also be a need to support ref kinds (not just invokevirtual).

nizarbenalla:
Based on some offline discussions, I made the function classes final with a private constructor.

PR #284, opened 2025-06-20
Generate Java enums

narangvivek10:
We are using jextract in [cuVS Java](https://github.com/rapidsai/cuvs/tree/branch-25.08/java). Right now, in jextract the C enums are generated as individual getter functions returning the ordinals. However, this results in code where it is difficult to disambiguate which functions are logically grouped together for a particular Enum. Additionally, given the lack of this capability, we have to manually create and maintain Java enums for each enum in the C layer.

In this PR, I wish to introduce an option to additionally generate Java enums by passing an optional flag `--generate-java-enums`. In case a Java user wishes to use bitwise operations on the enums, they can use the enum's `.ordinal()` method for this. The original static getter methods will still be generated as before. 

Example C enum:
```c
enum SIZE {
   S,
   M,
   L
};
```
Produces the following with jextract:
```java
private static final int S = (int)0L;
public static int S() {
     return S;
}

private static final int M = (int)1L;
public static int M() {
     return M;
}

private static final int L = (int)2L;
public static int L() {
     return L;
}
```
With this feature enabled, using the `--generate-java-enums`, the following is generated (in addition to the above):
```java
public enum Size {
    S(0),
    M(1),
    L(2);

    private final int value;

    private Size(int value) {;
        this.value = value;
    }

    public int getValue() {
        return this.value;
    }
}
```


mcimadamore:
In general we have not used the translation to Java enum because it's not 100% semantics preserving. In C enum constants are more of loose constants -- given:

```
enum Color
{
    Red,
    Green,
    Blue,
}
```

C code can do things like `Red | Blue` -- which are not possible with Java enums.

Moreover, C enum constants live in the same namespace as global variables. So it seemed again honest to allow clients to access enum constants directly after having imported (statically) the main generated header class.

While in some ways using Java enums to model C enums look more natural, it also represents a significant departure from C. I can imagine that some simple uses of C enums would be quite happy with the approach you propose -- but this approach would not scale for more complex uses.

The philosophy of jextract has, so far, been to generate a set of bindings that was as close as possible to the semantics of the extracted header. That has been the main guiding principle. There's a lot of "creativity" jextract could provide on top (and enums is a good example), but adding creativity almost always results in binding that work better in some cases and worse in others. For these reasons, I don't think this PR fits well with the design principles of jextract.

One less radical move could be to give up to the "single namespace" property, and at least group related enum constants together, under a separate generated class. That would mean the Java code would require more "qualifiers" than the associated C code, but with some better separation. Would something like that be enough? Or do you really need these constants to be enum constants (e.g. because you use them in exhaustive switches etc.) ?

mcimadamore:
> Hey @mcimadamore please excuse my gap in understanding here, but won't this operation be possible in Java by doing: `Color.Red.ordinal() | Color.Blue.ordinal()`?

As said in another comment, there's no guarantee that the Java enum ordinal and the value of the C enum constant will be the same. So, no, that would not be a good translation. Also, please consider clients passing enums to native functions -- right now they can just say e.g.

```java
native_func(RED);
```

It would be sad if they had to do:

```java
native_func(RED.ordinal());
```

Of course one could imagine generating one extra overload for each enum-accepting function -- but that has issues:

* you can only use the Java enum-accepting overload in some cases, but not other (where you want to use bitwise ops)
* if a function uses N enum types, you need 2^N overloads (at least if you want to generate all possible combinations)

I don't think much joy is to be had by modelling C enums with Java enums. We've been at it several times, and we keep retracing our steps back to the same answers.

mcimadamore:
> Thanks @mcimadamore for your feedback. Yes, if we can at least group related enum constants together under a separate generated class, that will be awesome. Would it be okay if I update this PR to have this change? Please let me know. Thanks!

I think we should ponder this some more. For simple use cases, it is really great to just have all the enum constants "in scope" in the same way as it happens with C. W/o that, the user would be left to try and figure out "what to include" for each enum. This might not be too obvious because the C source file one is trying to convert might only refer to enum _constants_ but never to the _enum type_ bringing these constants into existence. E.g. a C source might just use `Red` and `Green` and pass them to functions (and maybe these functions just take an `int`) -- and the code might never utter `Color`, ever. As a jextract user, how would I know that I need to `import static foo.Color.*`` to get there?

I'm not saying this is a blocker -- just listing all the various potential issues with changing the status quo (also as a way to explain why we are where we are).

mcimadamore:
> > What, in Java, gives you this OR-like functionality for enums? EnumSet!
> 
> Modelling each enum value as an enum set is certainly possible. It also sounds as overkill. One of the principles of jextract is to add as little overhead as possible. I think here we should just be honest and admit that C enums and Java enums don't have much in common, except for the use of the word `enum`.

(note -- this does NOT mean that other, more specialized, code generators could not model C enums as Java enums. Perhaps in some code bases that makes total sense. But jextract is intended to provide reliable bindings for _all_ C headers -- which sort of restricts the kind of tricks we can play)

Marcono1234:
> > Hey @mcimadamore please excuse my gap in understanding here, but won't this operation be possible in Java by doing: `Color.Red.ordinal() | Color.Blue.ordinal()`?
> 
> As said in another comment, there's no guarantee that the Java enum ordinal and the value of the C enum constant will be the same. So, no, that would not be a good translation. Also, please consider clients passing enums to native functions -- right now they can just say e.g.
> 
> ```java
> native_func(RED);
> ```

Isn't this what the enum `getValue()` method included in this PR is for? E.g. `Color.Red.getValue() | Color.Blue.getValue()` and `native_func(RED.getValue())`.

But the other concerns mentioned above still remain, such as making it difficult to discover which enum to use when a jextract-generated method just accepts an `int`, and that users might still accidentally use `ordinal()` or `MyEnum.values()[i]` instead of the `getValue()` method.

narangvivek10:
Apologies for the significant delay of several weeks on this. I will revisit this feature and the suggestions/feedback and will create a fresh PR (if needed) in the near future. Thanks!

kingg22:
Hi @kingg22, thanks for making a comment in an OpenJDK project!

All comments and discussions in the OpenJDK Community must be made available under the OpenJDK [Terms of Use](https://openjdk.java.net/legal/tou/). If you already are an OpenJDK [Author](https://openjdk.java.net/bylaws#author), [Committer](https://openjdk.java.net/bylaws#committer) or [Reviewer](https://openjdk.java.net/bylaws#reviewer), please click [here](https://bugs.openjdk.java.net/secure/CreateIssue.jspa?pid=11300&issuetype=1) to open a new issue so that we can record that fact. Please Use "Add GitHub user kingg22" for the summary.

If you are not an OpenJDK Author, Committer or Reviewer, simply check the box below to accept the OpenJDK Terms of Use for your comments.

 - [ ]  I agree to the [OpenJDK Terms of Use](https://openjdk.java.net/legal/tou/) for all comments I make in a project in the [OpenJDK GitHub organization](https://github.com/openjdk).

Your comment will be automatically restored once you have accepted the OpenJDK [Terms of Use](https://openjdk.java.net/legal/tou/).

PR #287, opened 2025-07-23
7903947: Access to function pointers  in structs could be streamlined

nizarbenalla:
Please review this patch to add a new option where you can generate an direct invoker methods. This feature is intended to be used on structs that are just a collection of functions we had like to call.

If this feature is used, we remove the getter to avoid name collisions as we assume this is similar to an interface.

Changes to `GUIDE.md` have been intentionally left out from this initial patch to make reviews easier.

```C
struct Foo {
    struct Bar (*a)(void);
    struct Bar (*b)(int);
};
```

Before 

```java
var foo = alloc_callback_h.foo();

var barA = Foo.a.invoke(Foo.a(foo), arena);
var barB = Foo.b.invoke(Foo.b(foo), arena, 100);
```

After 

```java
var foo = alloc_callback_h.foo();

var barA = Foo.a(foo, arena);
var barB = Foo.b(foo, arena, 100);

```

---------
### Progress

### Issue
 * [CODETOOLS-7903947](https://bugs.openjdk.org/browse/CODETOOLS-7903947): Access to function pointers  in structs could be streamlined (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/287/head:pull/287` \
`$ git checkout pull/287`

`$ git checkout pull/287` \
`$ git pull https://git.openjdk.org/jextract.git pull/287/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 287`

`$ git pr show -t 287`

</details>

<a href="https://git.openjdk.org/jextract/pull/287.diff">https://git.openjdk.org/jextract/pull/287.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/287#issuecomment-3107936946)
</details>

### DIFF
--- a/src/main/java/org/openjdk/jextract/impl/IncludeHelper.java
+++ b/src/main/java/org/openjdk/jextract/impl/IncludeHelper.java
@@ -33,7 +33,9 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
 import java.util.Comparator;
+import java.util.Collections;
 import java.util.EnumMap;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -82,12 +84,35 @@ public class IncludeHelper {
     }
 
     private final EnumMap<IncludeKind, Set<String>> includesSymbolNamesByKind = new EnumMap<>(IncludeKind.class);
+
+    private final Map<String, Set<String>> propertiesByName = new HashMap<>();
+
     private final Set<Declaration> usedDeclarations = new HashSet<>();
     public String dumpIncludesFile;
 
+    /**
+     * Register an include, may be:Name or Name,flag1,flag2 orName,prop=value,flag
+     */
     public void addSymbol(IncludeKind kind, String symbolName) {
-        Set<String> names = includesSymbolNamesByKind.computeIfAbsent(kind, (_unused) -> new HashSet<>());
-        names.add(symbolName);
+        String[] parts = symbolName.split(",", 2);
+        String name = parts[0].trim();
+
+        includesSymbolNamesByKind.computeIfAbsent(kind, k -> new HashSet<>())
+                .add(name);
+
+        if (parts.length == 2) {
+            Set<String> props = propertiesByName
+                    .computeIfAbsent(name, k -> new HashSet<>());
+            for (String tok : parts[1].split(",")) {
+                props.add(tok.trim());
+            }
+        }
+
+        if (kind == IncludeKind.STRUCT || kind == IncludeKind.UNION) {
+            includesSymbolNamesByKind
+                    .computeIfAbsent(IncludeKind.TYPEDEF, k -> new HashSet<>())
+                    .add(name);
+        }
     }
 
     public boolean isIncluded(Declaration.Variable variable) {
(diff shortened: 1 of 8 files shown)


mcimadamore:
Just to recap solution we have explored previously.

It would be nice if, in principle, just doing this:

```
var bar = Foo.a(foo, <args>);
```

worked, and it invoked the correct function pointer on the `foo` struct. But this solution can lead to clashes. Specifically, if the function pointer takes no arguments, then `Foo.a(foo)` would look exactly like a struct getter.

Another solution would be to tweak `a::invoke` method to work on a `foo` instance directly, and not on the _address_ that function takes. But this breaks composition -- if you first obtain the function pointer (a MemorySegment) stored at `Foo::a`, and then want to invoke it at a later point, how would you do it? If `invoke` takes a segment for `Foo` you can't really -- you always need to _first_ store the address in the `Foo` struct, and _then_ call the desired function pointer using `a::invoke`. Which seems awkward.

(note also that we cannot "just" add an overload of `a::invoke` that takes a Foo, because that is a memory segment too, so we get another clash).

In other words, there's no "easy" solution here. One possibility might be to add an invoker accessor to `Foo` -- something like `Foo.a$invoke(<args>)`. Since this uses a different name there would be no clash with the getter. This leaves the problem that, in structs that contain only function pointers, the getters are probably not very useful, but maybe we can add logic to filter those out.

The other solution is the one explored here: the developer _knows_ this isn't just _any_ struct, but a "functional" struct. It tells jextract, and jextract reacts by generating slightly different code. Perhaps one issue with this approach (although we arrived at it honestly) is that if one stares at code like:

```
var barA = Foo.a(<args>);
```

It is pretty difficult to tell whether that's a getter/setter or an "invoker".

src/main/java/org/openjdk/jextract/impl/StructBuilder.java
```diff
@@ -137,25 +142,102 @@ public void addVar(Declaration.Variable varTree) {
...
                 emitFieldArrayGetter(javaName, varTree, arrayHandle, indexList);
                 emitFieldArraySetter(javaName, varTree, arrayHandle, indexList);
             }
-        } else if (Utils.isPointer(varTree.type()) || Utils.isPrimitive(varTree.type())) {
-            emitFieldGetter(javaName, varTree, layoutField, offsetField);
-            emitFieldSetter(javaName, varTree, layoutField, offsetField);
+        } else if (Utils.isPointer(type) || Utils.isPrimitive(type)) {
+            boolean isFuncPtr = functionalDispatch && Utils.isFunctionPointer(type);
+            if (isFuncPtr) {
+                emitFieldSetter(javaName, varTree, layoutField, offsetField);
```
mcimadamore, line 160: I'm not sure we even want the setter -- the fields in these structs are typically populated by a library

src/main/java/org/openjdk/jextract/impl/IncludeHelper.java
```diff
@@ -82,12 +84,35 @@ static IncludeKind fromScoped(Declaration.Scoped scoped) {
...
+
+        if (parts.length == 2) {
+            Set<String> props = propertiesByName
+                    .computeIfAbsent(name, k -> new HashSet<>());
+            for (String tok : parts[1].split(",")) {
+                props.add(tok.trim());
+            }
+        }
+
+        if (kind == IncludeKind.STRUCT || kind == IncludeKind.UNION) {
```
mcimadamore, line 111: Not sure I get this logic -- it seems like we're always setting a property P for a typedef with name N whenever we see it set for a struct N? That seems odd.

But in example as the one you added in the test are a bit ambiguous, as the same declaration induces both an anonymous struct decl and a typedef, so it is less clear which is which. If I write:

```
typedef struct Foo {
int a;
};
```

It seems to be that jextract only retains a "struct" tree, and throws away the typedef. In fact I was surprised that your test worked at all, as in the above case `dump-includes` would include something like this:

```
--include-struct Foo   # header: ....
```

And *not*:

```
--include-typedef Foo   # header: ....
```

So adding a property on the typedef should have no effect?

mcimadamore:
> Will the use of ",functional" be the intended way to trigger this
> enhancement? I wonder whether last error capture could be built in with
> "--include-function funcname,erro" or "--include-function
> funcname,GetLastError"

Nice connection -- we do in fact have plans (at least in theory) to use this same mechanism for errno -- and maybe even to declare "disjoint enums" (e.g. enums that can be modelled as plain Java enums because you swear not to rely on low-level bitmask operation -- which is often the case if the enum you are modelling is just a "tag")

mcimadamore:
> This proposal makes it nicer to handle invokers. I generated my code with
> this change but as you'vr noted it does indeed lead to clashes using the
> proposed naming convention.

I think the issue is that the changes in this PR generate both setter and invoker -- and we should only emit the invoker.

nizarbenalla:
I kept trying to avoid renaming the invoker as I didn't like having the dollar sign or a prefix in the invoker, but I like the idea of having readonly/write/execute mode in the config.

mcimadamore:
> > I kept trying to avoid renaming the invoker as I didn't like having the dollar sign or a prefix in the invoker, but I like the idea of having readonly/write/execute mode in the config.
> 
> I feel your pain. We defo went thorugh a lot of pain to remove dollars from generated code:
> 
> https://cr.openjdk.org/~mcimadamore/panama/jextract_changes.html
> 
> So, it will be a bit sad to see them coming back. Also, it will make some code more verbose compared to their C counterpart, as the C syntax can distinguish between application (`foo.x()`) and access (`foo.x = ...`).

Maybe worth thinking some more about what our options are...

PR #288, opened 2025-09-03
7904073: Derive required symbols from client code to minimize generated bindings

nizarbenalla:
Please review this change to add a small sample that extracts full bindings once, scans the compiled client classes to collect actually referenced symbols, filters the includes list, and re-runs jextract to produce a slim binding set.

### DIFF
--- a/samples/opengl-minimal-bidings/Teapot.java
+++ b/samples/opengl-minimal-bidings/Teapot.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright (c) 2025, Oracle and/or its affiliates. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *   - Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *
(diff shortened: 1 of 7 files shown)


samples/opengl-minimal-bidings/Teapot.java
```diff
@@ -0,0 +1,88 @@
...
+
+    void onIdle() {
+        rot += 0.1;
+        glutPostRedisplay();
+    }
+
+    public static void main(String[] args) {
+        try (var arena = Arena.ofConfined()) {
+            var argc = arena.allocateFrom(C_INT, 0);
+            glutInit(argc, argc);
```
JornVernee, line 76: Second argument shouldn't be `argc`, bu a pointer to `argv` (https://www.opengl.org/resources/libraries/glut/spec3/node10.html). 

I think the other sample also gets this wrong, but I think it works out because we pass `0` as `argc`.

nizarbenalla:
@keithc-ca I cannot read and respond to your comment, you will need to accept the OpenJDK Terms of Use for your comment to appear.

nizarbenalla:
I cannot read your comment because you have not accepted the OpenJDK TOU.

PR #300, opened 2026-02-04
7904128: Add a struct reinterpret helper that keeps scope

nizarbenalla:
Please review this change to add a new reinterpret method that keeps scope, I have removed one of the old methods and only kept one with arena and cleanup, and one without.

---------
### Progress

### Issue
 * [CODETOOLS-7904128](https://bugs.openjdk.org/browse/CODETOOLS-7904128): Add a struct reinterpret helper that keeps scope (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/300/head:pull/300` \
`$ git checkout pull/300`

`$ git checkout pull/300` \
`$ git pull https://git.openjdk.org/jextract.git pull/300/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 300`

`$ git pr show -t 300`

</details>

<a href="https://git.openjdk.org/jextract/pull/300.diff">https://git.openjdk.org/jextract/pull/300.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/300#issuecomment-3848490341)
</details>

### DIFF
--- a/doc/GUIDE.md
+++ b/doc/GUIDE.md
@@ -328,10 +328,9 @@ public class Point {
     public static MemorySegment allocateArray(long elementCount,
             SegmentAllocator allocator) { ... } // 4
 
-    public static MemorySegment reinterpret(MemorySegment addr, Arena arena,
-            Consumer<MemorySegment> cleanup) { ... } // 6
     public static MemorySegment reinterpret(MemorySegment addr, long elementCount,
             Arena arena, Consumer<MemorySegment> cleanup) { ... } // 6
+    public static MemorySegment reinterpret(MemorySegment addr, long elementCount) { ... } // 6
 }
 ```
 
@@ -406,17 +405,24 @@ The `reinterpret` method can be used to associate the correct bounds and lifetim
 // Main.java
 
 try (Arena arena = Arena.ofConfined()) {
-    MemorySegment point = Point.reinterpret(new_point(), arena, mylib_h::delete_point);
+    MemorySegment point = Point.reinterpret(new_point(), 1, arena, mylib_h::delete_point);
 
     // ...
 } // 'delete_point` called here
 ```
 
 The `point` segment returned by `reinterpret` has exactly the size of one `Point` (for
-arrays of struct, use the `reinterpret` overload that takes an element count as well). The
-lifetime we associate with the segment is the lifetime denoted by `arena`, and when the
-arena is closed, we want to call `delete_point`, which we can do by passing a method
-reference to `delete_point` as a cleanup action when calling `reinterpret`.
+arrays of struct, pass a larger `elementCount`). The lifetime we associate with the segment
+is the lifetime denoted by `arena`, and when the arena is closed, we want to call
+`delete_point`, which we can do by passing a method reference to `delete_point` as a cleanup
+action when calling `reinterpret`.
+
+If the source segment already has the desired scope and you only need to adjust the bounds, use
+the overload that keeps the existing scope:
+
+```java
+MemorySegment points = Point.reinterpret(existingSegment, elementCount);
+```
 
 The class that jextract generates for unions is identical to the class generated for
 structs.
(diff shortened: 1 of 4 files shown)


mcimadamore (commented):
In general, the change seems good. I'm a bit unsure re. compatibility impact -- the test changes shows that there's some code out there that might stop working. If we want to play extra safe we could keep all existing methods, and add a new one.

nizarbenalla:
I suggest we go ahead with this change, as I could find a limited number of call sites of the removed method when looking at public code.

JornVernee:
Why are we removing the overload for a single element though? Can't we just add two more overloads that don't accept an arena/cleanup action, or maybe allow users to pass `null` as an arena, and then select the right overload in the impl?

mcimadamore:
So, old one was:

1. elemCount, arena, cleanup
2. arena, cleanup

New one is:

3. elemCount
1. elemCount, arena, cleanup

So, if I'm right there's _three_ shapes, not _four_ ? Can we have all three?

JornVernee:
I think the fourth shape just takes a MemorySegment, and then reinterprets it to have the size of the struct (but leaves the scope the same).

mcimadamore:
> I think the fourth shape just takes a MemorySegment, and then reinterprets it to have the size of the struct (but leaves the scope the same).

Yes, but we never provided that, right?

JornVernee:
> > I think the fourth shape just takes a MemorySegment, and then reinterprets it to have the size of the struct (but leaves the scope the same).
> 
> Yes, but we never provided that, right?

That's right. I think we just wanted two overloads: one for single structs, and one for arrays.

mcimadamore:
> > > I think the fourth shape just takes a MemorySegment, and then reinterprets it to have the size of the struct (but leaves the scope the same).
> > 
> > 
> > Yes, but we never provided that, right?
> 
> That's right. I think we just wanted two overloads: one for single structs, and one for arrays.

So, to be clear, the full matrix is:

1. no-arg
2. arena, cleanup
3. element count
4. element count, arena, cleanup

I mean -- having all four does seem a bit over the top. And I think of all these, I have a feeling (2) seems the least useful.

Also, (1) is easily subsumed by (3) -- which is I think what @nizarbenalla is aiming for with this PR.

So, if we wouldn't mind about compatibility, I think I would agree that having (3) and (4) is probably enough, and users should just pass `1` for a single element.

But we do have some compatibility constraints with (2) -- IMHO, (2) is probably rarely used, and the workaround is simple, so maybe that's ok?

JornVernee:
> IMHO, (2) is probably rarely used

I've found myself using this overload pretty much always when using jextract bindings to get automatic cleanup (through whatever "free" function the library offers). e.g. a library has some function that returns a 'handle' to same library-related state, and I use `reinterpret` to attach it to an arena with a cleanup action that calls the relevant 'close'/'free' function for the handle. But, most of the time I'm working on a pretty simple program. That said, adding the extra `1` doesn't seem like that much effort.

mcimadamore:
> > IMHO, (2) is probably rarely used
> 
> I've found myself using this overload pretty much always when using jextract bindings to get automatic cleanup (through whatever "free" function the library offers). e.g. a library has some function that returns a 'handle' to same library-related state, and I use `reinterpret` to attach it to an arena with a cleanup action that calls the relevant 'close'/'free' function for the handle. But, most of the time I'm working on a pretty simple program. That said, adding the extra `1` doesn't seem like that much effort.

Uhm -- true. In my experience, attaching cleanup actions is most useful with opaque pointers (e.g. createFoo, destroyFoo) where in most cases you don't have accessors directly, and you access through the library itself. But perhaps I oversampled based on the libraries I worked with.

mcimadamore:
> That said, adding the extra `1` doesn't seem like that much effort.

(It's not about effort, I think. It's about 2x-ing reinterpret methods, when some might feel generated code is already too verbose)

mcimadamore:
> > That said, adding the extra `1` doesn't seem like that much effort.
> 
> (It's not about effort, I think. It's about 2x-ing reinterpret methods, when some might feel generated code is already too verbose)

E.g. -- there's no way to filter out these, unless you start messing with generated code.

mcimadamore:
> > > That said, adding the extra `1` doesn't seem like that much effort.
> > 
> > 
> > (It's not about effort, I think. It's about 2x-ing reinterpret methods, when some might feel generated code is already too verbose)
> 
> E.g. -- there's no way to filter out these, unless you start messing with generated code.

That, said, I understand that having 3 instead of 4 is not a particularly principled position either. I don't really have a strong opinion about all this.

PR #301, opened 2026-02-23
Print IncludeKinds in consistent order

yuzawa-san:
- the current jextract `--dump-includes` output inconsistently orders symbols by include type. (e.g. struct, typedef, function)
- this means that when i rerun jextract dump includes on newer version header files, the output is not in stable order. this causes extraneous and confusing diffs. i believe this is a caused of the fact that Collectors.groupingBy uses some sort of HashMap which is finicky around ordering.
- solution is to use a EnumMap which naturally keeps the ordering of the IncludeKinds
- i was not able to get all of the testing working (even on master) locally, but the salient test `testDumpIncludes` passes.

---------
### Progress

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/301/head:pull/301` \
`$ git checkout pull/301`

`$ git checkout pull/301` \
`$ git pull https://git.openjdk.org/jextract.git pull/301/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 301`

`$ git pr show -t 301`

</details>

<a href="https://git.openjdk.org/jextract/pull/301.diff">https://git.openjdk.org/jextract/pull/301.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/301#issuecomment-3942202842)
</details>

### DIFF
--- a/src/main/java/org/openjdk/jextract/impl/IncludeHelper.java
+++ b/src/main/java/org/openjdk/jextract/impl/IncludeHelper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2021, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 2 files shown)


JornVernee:
Good change!

In the same code we are already using TreeMap and TreeSet to make the order of the declarations themselves deterministic:

```
Map<Path, Set<Declaration>> declsByPath = usedDeclarations.stream()
        .collect(Collectors.groupingBy(d -> d.pos().path(),
                () -> new TreeMap<>(Path::compareTo),
                Collectors.toCollection(() -> new TreeSet<>(Comparator.comparing(Declaration::name)))));
```

Looks like we just missed the include kinds.

I've ran the tests locally, and they all pass. But, it would be nice to add an extra test that checks that the output is deterministic.

JornVernee (approved):
Good

JornVernee:
@yuzawa-san To continue to merge this, you will need to issue the `/integrate` command. (see https://github.com/openjdk/jextract/pull/301#issuecomment-3942197720)

PR #308, opened 2026-07-31
Add an option for copying comments in a header file over to the generated bindings

xpple:
This PR adds an option for copying comments in a header file over to the generated bindings. See also [my previous email](https://mail.openjdk.org/archives/list/jextract-dev@openjdk.org/thread/R3DOKMJVNOGYHH22H5F3ABS67UGWM4IB/).

The option can be enabled by passing the `--copy-comments` flag to jextract. This flag is optional, and false by default. This means that if the flag is not passed, the output from jextract will be identical to what it was before this change.

I decided against using the existing clang API for Doxygen comments, because they are far from standardized in C projects. This meant though that I had to come up with custom logic for associating comments with declarations. I am happy to incorporate feedback on how this association can be improved. Currently, all comments that immediately precede a declaration are associated with that declaration. Whitespace is ignored.

For example, in the below code, both comments are associated with the `func` declaration.

```c
// one
// two
void func(void);
```

For structs/unions, comments that precede member declarations will be associated with that member.

```c
/**
 * will be associated with Point
 */
struct Point {
    /// will be associated
    /// with x
    int x;
    /// will be associated
    /// with y
    int y;
};
```

Similarly, enum constants can have comments too. Comments at the enum tag are ignored, because the enum tag itself is not retained in the jextract bindings.

```c
// this comment is ignored
enum Color {
    // will be associated with RED
    RED,
    // will be associated with GREEN
    GREEN,
    // will be associated with BLUE
    BLUE,
};
```

In the API, the `Declaration` interface now has a `List<String> comments()` method. These are the raw comments, so they still contain the comment delimiters (`//` or `/* ... */`). If `--copy-comments` is not passed, this method will return an empty list (not `null`).

In the binding generation code, I made an attempt to detect some common styles for comments, and account for them separately. For example, if a comment starts with three forward slashes, all three will be stripped (and not just the two that lexically define the comment). The code for this is in `ClassSourceBuilder#copyComments`. Below is what the JavaDoc will look like for the above three examples.

<details><summary>func</summary>

```java
/**
 * {@snippet lang=c :
 * void func()
 * }
 * <p><strong>Copied comments:</strong></p>
 * one
 * two
 */
public static void func() {
    // ...
}
```

</details>

<details><summary>Color</summary>

```java
/**
 * {@snippet lang=c :
 * enum Color.RED = 0
 * }
 * <p><strong>Copied comments:</strong></p>
 * will be associated with RED
 */
public static int RED() {
    // ...
}

/**
 * {@snippet lang=c :
 * enum Color.GREEN = 1
 * }
 * <p><strong>Copied comments:</strong></p>
 * will be associated with GREEN
 */
public static int GREEN() {
    // ...
}

/**
 * {@snippet lang=c :
 * enum Color.BLUE = 2
 * }
 * <p><strong>Copied comments:</strong></p>
 * will be associated with BLUE
 */
public static int BLUE() {
    // ...
}
```

</details>

<details><summary>Point</summary>

```java
/**
 * {@snippet lang=c :
 * struct Point {
 *     int x;
 *     int y;
 * }
 * }
 * <p><strong>Copied comments:</strong></p>
 * will be associated with Point
 */
public class Point {
    // ...

    /**
     * Getter for field:
     * {@snippet lang=c :
     * int x
     * }
     * <p><strong>Copied comments:</strong></p>
     * will be associated
     * with x
     */
    public static int x(MemorySegment struct) {
        // ...
    }

    /**
     * Setter for field:
     * {@snippet lang=c :
     * int x
     * }
     * <p><strong>Copied comments:</strong></p>
     * will be associated
     * with x
     */
    public static void x(MemorySegment struct, int fieldValue) {
        // ...
    }

    // ...

    /**
     * Getter for field:
     * {@snippet lang=c :
     * int y
     * }
     * <p><strong>Copied comments:</strong></p>
     * will be associated
     * with y
     */
    public static int y(MemorySegment struct) {
        // ...
    }

    /**
     * Setter for field:
     * {@snippet lang=c :
     * int y
     * }
     * <p><strong>Copied comments:</strong></p>
     * will be associated
     * with y
     */
    public static void y(MemorySegment struct, int fieldValue) {
        // ...
    }

    // ...
}
```

</details>

Not all JavaDoc comments pertaining to a certain declaration get the comments associated with that declaration. For example, for functions the `$descriptor`, `$handle` and `$address` methods do not get the comments, and for structs the `$layout` and `$offset` methods do not get them. The documentation is usually only relevant for calling/accessing/modifying something, not for derived properties.

The comment association code works by keeping track of the previous cursor's extent's ending, and tokenizing between that and the current cursor's extent's beginning. For example, if in the below code the current cursor is `func2`, then the tokens encountered will be `;`, `// comment` and `void`.

```c
void func1(void);

// comment
void func2(void);
```

These tokens will be iterated over in reverse order, ignoring the first token(s) (here only `void`) as they are part of the current declaration. Then, as long as the encountered token is a comment (`CXToken_Comment`), it will be associated with the declaration. In cases where this approach can't work (e.g. there is no previous declaration, the previous declaration is in a different file or the previous declaration actually appears after the current declaration lexically), the code will resort to a fallback.

Since I needed `clang_getRange`, which wasn't in the symbols list (`updateclang/clang.symbols`) yet, I had to regenerate the bindings. Because jextract had received some updates since the last time the bindings were generated, there are some bindings changes unrelated to `clang_getRange` as well. While updating, I noticed the instructions had gotten slightly out of date, so I updated those as well. I added three patch files which people can `git apply` to apply the necessary changes more easily.

I made sure to maintain source and binary compatibility in the public API of jextract.

I did not test the PR _locally_, but the tests I added passed on GitHub Actions. As this is my first time working with libclang, I am happy receive feedback for this PR.

- [x] I confirm that I make this contribution in accordance with the [OpenJDK Interim AI Policy](https://openjdk.org/legal/ai).

---------
### Progress
- \[x\] Change must not contain extraneous whitespace
- \[x\] Change must be properly reviewed (no review required)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jextract.git pull/308/head:pull/308` \
`$ git checkout pull/308`

`$ git checkout pull/308` \
`$ git pull https://git.openjdk.org/jextract.git pull/308/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 308`

`$ git pr show -t 308`

</details>

<a href="https://git.openjdk.org/jextract/pull/308.diff">https://git.openjdk.org/jextract/pull/308.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jextract/pull/308#issuecomment-5146921585)
</details>


xpple:
There is currently a bug where copied comments could contain `*/`, ending the emitted JavaDoc comment and resulting in invalid Java code. This can be fixed by removing `*/` from the copied comments[^1], but I think a better fix is to use Markdown comments instead. I have [code ready locally](https://github.com/xpple/jextract/commit/488a490df21f117a709256230184960679df1dd1) that migrates jextract to use Markdown comments everywhere, but I'll hold back on pushing that code because this migration is probably something that should be discussed first.

For what it's worth, I think using Markdown comments is a great idea. It would break compatibility with JDK 22, but I don't think retaining that compatibility is a goal.

[^1]: Note that trivial replacements fall victim to the classic nonidempotent sanitisation problem. E.g. the comment `/// **//` would still contain `*/` after removing `*/`.

xpple:
Please do! I believe everything to be working currently, but it's possible I missed something.

src/main/java/org/openjdk/jextract/impl/DeclarationImpl.java
```diff
@@ -42,6 +43,7 @@ public abstract class DeclarationImpl implements Declaration {
 
     private final String name;
     private final Position pos;
+    private List<String> comments = Collections.emptyList();
```
mcimadamore: This could (and should?) be modeled as another attribute class.

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
```diff
@@ -292,6 +308,43 @@ static String declarationComment(Declaration decl) {
         return declString.lines().collect(Collectors.joining("\n * ", " * ", ""));
     }
 
+    static String copyComments(Declaration decl) {
+        List<String> comments = decl.comments();
+        if (comments.isEmpty()) {
+            return "";
+        }
+
+        return " * <p><strong>Copied comments:</strong></p>\n" + comments.stream()
```
mcimadamore, line 319: Not sure whether we need the header -- but perhaps this is similar to "inherited doc" :-)

test/testng/org/openjdk/jextract/test/toolprovider/copiedComments/functions.h
```diff
@@ -0,0 +1,72 @@
...
+/*
+ * internal
+ *
+ * comment
+ */
+void func7(int a);
+
+
+/**
+ * Doxygen
```
mcimadamore, line 67: In this case we also copy "Doxygen" right? I'm a bit unsure here that the copying will work correctly as some of the Doxygen tags will confuse javadoc (e.g. `@param`)

mcimadamore (commented):
Overall I like the pragmatic approach of looking at tokens and opportunistically copy comments over. We need to make sure to escape any javadoc sensitive string -- like `@`, `*/` or html tags -- e.g. you really want to treat what's been copied as "opaque text" (and maybe beef up the tests in this direction)

Re. doxygen, while I agree with your conclusion that, alone, is not enough -- I think a natural follow up would be to add direct support for it -- which will result in more pleasing output.

Finally, I'd suggest to split the re-generation of the clang files -- plus your patch publication in a separate PR so as not to accumulate too much noise.

xpple:
Thanks for the review! What is your stance on using Markdown comments for jextract (see [this comment](https://github.com/openjdk/jextract/pull/308#issuecomment-5160329039))?

src/main/java/org/openjdk/jextract/impl/ClassSourceBuilder.java
xpple: Yeah. It was mostly to distinguish it from the existing JavaDoc.

xpple:
Regarding escaping special JavaDoc characters, I suppose that is a design choice/trade-off. Right now some Doxygen tags kind of "just work", in an implicit way. For example

```c
/**
 * Foo.
 *
 * @param size size
 * @param data data
 * @return success code
 * @retval 1 if successful
 * @retval 0 if unsuccessful
 */
int foo(int size, int data[size]);
```

produces the following JavaDoc:

```java
/**
 * {@snippet lang=c :
 * int foo(int size, int data[size])
 * }
 * <p><strong>Copied comments:</strong></p>
 * Foo.
 * 
 * @param size size
 * @param data data
 * @return success code
 * @retval 1 if successful
 * @retval 0 if unsuccessful
 */
```

My IDE renders this as:

image

Note that `@retval` is not a standard JavaDoc tag. As such, the official JavaDoc tool ignores it, unless you define the custom tag yourself:

image

I think it is nice that Doxygen tags that share the same syntax as JavaDoc tags maintain functionality. Perhaps we can make this transcription explicit rather than implicit. I suppose if we use the clang API for associating comments, the clang API can also help us here.

For non-Doxygen comments, I agree that we should escape all special JavaDoc characters.

Should differentiating between Doxygen comments and non-Doxygen comments be part of this PR, or a future PR?

mcimadamore:
> Thanks for the review! What is your stance on using Markdown comments for jextract (see [this comment](https://github.com/openjdk/jextract/pull/308#issuecomment-5160329039))?

I don't have a strong opinion, but it's a separate effort. I tend to agree that jextract doesn't have big compatibility commitment, so there's room to change things

mcimadamore:
> I suppose if we use the clang API for associating comments, the clang API can also help us here.

yep -- that's what I had in mind.

I'm not convinced the "accidentally works" is a good property here :-)

xpple:
Okay, let's first support copying the comments as opaque text, and later add support for Doxygen comments. However, I have two questions.

1. Should we keep the following "normalisation" rules?
   - If a comment starts with `///`, strip all three slashes and not just the two lexically defining the comment. As this is a Doxygen style comment, which we plan to support later separately, this technically violates the opaque text interpretation.
   - If a line in a multi-line comment starts with `* `, should we strip it? Currently I do
     ```java
     .collect(Collectors.joining("\n")).lines()
     .map(String::strip)
     .map(line -> {
         if (line.startsWith("* ")) {
             return line.substring("* ".length());
         }
         if (line.startsWith("*")) {
             return line.substring("*".length());
         }
         return line;
     })
     .map(line -> " * " + line + "\n")
     ```
     If we interpret the comment strictly as opaque text you could get a line with ` * * <rest of comment>`. 
2. How should I escape special JavaDoc characters? Quoting the [Escape Sequences](https://docs.oracle.com/en/java/javase/25/docs/specs/javadoc/doc-comment-spec.html#escape-sequences) section of the JavaDoc specification:
   
   > Escape sequences are context-sensitive, and can only be used where the use of the escaped character by itself would have a different syntactic interpretation. In other situations, these character sequences are taken literally, without additional interpretation.
   
   It seems nontrivial to do this correctly. As an example, you can't just replace all `@` with `@@`. Maybe the `jdk.javadoc` module can help? Not sure.

mcimadamore:
> Okay, let's first support copying the comments as opaque text, and later add support for Doxygen comments. However, I have two questions.
> 
>     1. Should we keep the following "normalisation" rules?
>        
>        * If a comment starts with `///`, strip all three slashes and not just the two lexically defining the comment. As this is a Doxygen style comment, which we plan to support later separately, this technically violates the opaque text interpretation.
>        * If a line in a multi-line comment starts with `* `, should we strip it? Currently I do
>          ```java
>          .collect(Collectors.joining("\n")).lines()
>          .map(String::strip)
>          .map(line -> {
>              if (line.startsWith("* ")) {
>                  return line.substring("* ".length());
>              }
>              if (line.startsWith("*")) {
>                  return line.substring("*".length());
>              }
>              return line;
>          })
>          .map(line -> " * " + line + "\n")
>          ```
>          
>          
>              
>                
>              
>          
>                
>              
>          
>              
>            
>          If we interpret the comment strictly as opaque text you could get a line with ` * * <rest of comment>`.
> 
>     2. How should I escape special JavaDoc characters? Quoting the [Escape Sequences](https://docs.oracle.com/en/java/javase/25/docs/specs/javadoc/doc-comment-spec.html#escape-sequences) section of the JavaDoc specification:
>        > Escape sequences are context-sensitive, and can only be used where the use of the escaped character by itself would have a different syntactic interpretation. In other situations, these character sequences are taken literally, without additional interpretation.
>        
>        
>        It seems nontrivial to do this correctly. As an example, you can't just replace all `@` with `@@`. Maybe the `jdk.javadoc` module can help? Not sure.

Good questions. Re (1) I think swallowing the beginning of a comment is fine, for both `///` and `*`.
Re (2), I agree it's more complex than it seems. 

That said, I feel that escaping the following:

* `&`, `<`, `>` -- because HTML
* `@` because javadoc
* `\`  because unicode

Should be ok. Replacement should use corresponding HTML entities, so:

* `@`   → `&#64;`
* `\`   → `&#92;`

What do you think?

JornVernee:
Could we quote verbatim by embedding the copied comment inside a code block? It might not look as good, but then we don't have to worry about quoting individual characters.

xpple:
That could be an option. Note that you still have to do some sanitisation, though. If we go with
```java
/**
 * <pre>
 * {@code
 * %s
 * }
 * </pre>
 */
```
then you still have to escape at least `*/` and `}`.

xpple:
I have another point for discussion. Consider the following comment:

```c
/// this is a comment
/// that spans over
/// multiple lines
```

Currently, the emitted JavaDoc for this would be

```java
/**
 * this is a comment
 * that spans over
 * multiple lines
 */
```

In the rendered JavaDoc, this becomes

```
this is a comment that spans over multiple lines
```

Note that the comment correctly renders on one line. Now consider this comment:

```c
/// @brief A function.
/// @param input Input.
/// @return The result.
int func1(int input);
```

The emitted JavaDoc would be

```java
/**
 * &#64;brief A function.
 * &#64;param input Input.
 * &#64;return The result.
 */
```

Importantly, this renders as

```
@brief A function. @param input Input. @return The result.
```

That is, it also renders on one line. Because `&#64;` is not a special JavaDoc character (unlike `@`), the text gets concatenated like any other text. Should we do anything with this? I think I lean towards no, because we do not support Doxygen (yet), and so the sanitization is doing its job correctly. What do you think? The formatting issue would be resolved when we add Doxygen support.

JornVernee:
I think using code blocks would preserve the original format as well? For sanitizing: would it help if we switch to markdown javadoc comments? When using markdown, code blocks can be created with 4 space indentation, and I believe that will retain the content verbatim.

I did a quick test with a javadoc comment like this:

```
/// Testing code blocks
///
///     /// @brief A function.
///     /// @param input Input.
///     /// @return The result.
///     ```
///     int func1(int input);
```

And that yields the desired result:

image

xpple:
I had completely forgotten you can also create code blocks like that. This seems like the perfect solution to me. Since I already have some code to migrate jextract to use Markdown comments, shall I open a separate PR for that?

<details><summary>Note about editor support</summary>

I tried the syntax in my IDE (IntelliJ IDEA), but found that it didn't work. It would just treat the indented text as regular text, removing the spaces.

I searched through IntelliJ's code and found this in [JavaDocMarkerProcessorFactory.java](https://github.com/JetBrains/intellij-community/blob/971b65fa9fd20ab9d58b111343ea5d5f0a3ead77/java/java-impl/src/com/intellij/codeInsight/javadoc/markdown/JavaDocMarkerProcessorFactory.java#L30-L31).

> Marker processor taking account of changes brought to Common Markdown by the JEP-467
> Namely: no indented code blocks, tables from Github Markdown

I'm not sure where they got "no indented code blocks" from. I looked at the initial revision of JEP 467 and it explicitly mentions indented code blocks. I will open a bug report for this tomorrow.

image

I then tried it in VS Code, and noticed it also doesn't work there. I didn't dig through the code to find out why, but I'll also open a bug report.

image

IDE support is quite important here, as it's probably less common to read jextract bindings documentation in a browser.

</details>

mcimadamore:
> What do you think? The formatting issue would be resolved when we add Doxygen support.

I tend to agree with this. I wouldn't worry too much about what happens with Doxygen rendering, given that we plan to add better support for that later on.

mcimadamore:
> I think using code blocks would preserve the original format as well? For sanitizing: would it help if we switch to markdown javadoc comments? When using markdown, code blocks can be created with 4 space indentation, and I believe that will retain the content verbatim.

Code blocks can help (code snippets can probably help too). But if the suggestion is to render all imported comments that way, IMHO that leads to a worse UX even for comments not using any weird chars. The proposed plan to do some basic sanitization (what we have in the PR looks ok?) plus enabling more direct doxygen support later seems to provide a better experience?

JornVernee:
> Since I already have some code to migrate jextract to use Markdown comments, shall I open a separate PR for that?

Yes, I think it's probably better to do that separately (again to filter out the noise). Jextract has always tried to be an early adopter of new JDK features, so using markdown comments seems like a good idea for that reason as well.

> IDE support is quite important here, as it's probably less common to read jextract bindings documentation in a browser.

I agree with you there. The picture I posted was from the html page generated by javadoc, but I think most people will indeed by looking at the docs in their IDE.

Thanks for filing bug reports! I can reproduce the issue in IntelliJ as well.

xpple:
Regarding code blocks vs. sanitization, clearly both have their pros and cons. Maybe we can ask users of jextract for opinions/feedback?

Regarding the IDE bugs, I have since filed the bug reports:

- https://youtrack.jetbrains.com/issue/IDEA-393352/Markdown-documentation-comments-indented-code-blocks-render-as-plain-text
- https://github.com/microsoft/vscode/issues/333770

JornVernee:
> if the suggestion is to render all imported comments that way, IMHO that leads to a worse UX even for comments not using any weird chars.

IMO it depends on the contents of the comment. If the comment is just some text, it's probably fine if that is re-flowed in the rendered javadoc. If the comment contains content that is intended to be in a certain format, then the re-flow will destroy this.

That said, I think the most common content of code comments is probably unformatted test (judging based on the JDK docs we write), so on the balance you're probably right that just doing some sanitizing will yield better results.

xpple:
> That said, I think the most common content of code comments is probably unformatted test (judging based on the JDK docs we write), so on the balance you're probably right that just doing some sanitizing will yield better results.

Okay, I think the PR is mostly finished then[^1]. I have added an additional test for a comment with multiple block tags that verifies our intended design (i.e. no line breaks or anything similar that would keep the content on separate lines as it was in the Doxygen comment).

A decision that still needs to be made is whether we want to keep the "**Copied comments**:" header. I like it because it distinctly separates the documentation of binding itself (text like `Getter for variable:`, and the C snippet) and the documentation corresponding to the native function.

Also I would like to request feedback on my strategy here:

https://github.com/openjdk/jextract/blob/976b708249f7e3bcaaf4c2bd43037e03d5f6ab55/src/main/java/org/openjdk/jextract/impl/TreeMaker.java#L146-L152

For example, for an opaque pointer like

```c
// OPAQUE_PTR comment
typedef struct Foo* OPAQUE_PTR;
```

the cursor `struct Foo` comes _before_ (in AST order) the cursor `typedef struct Foo* OPAQUE_PTR`. As such, when we want to find the comment tokens that precede `OPAQUE_PTR`, we can't use the ending of the previous cursor's extent as starting point for the tokenization. And so currently my fallback is to just use the start of the file as starting point. If multiple such typedefs are present, this will cause repeated tokenization of the code before the typedefs. This is not optimal, but at least in my testing it is also not really problematic. I'm also not quite sure what a good fix would be without significantly complicating the code.

---

In the meantime, I've been working on Doxygen support. I'll open a PR for that when this one is properly finished. You can see my code here: [copy-comments..doxygen-comments](https://github.com/xpple/jextract/compare/copy-comments..doxygen-comments).

[^1]: When #311 is merged, the PR should become a little cleaner for review.

xpple:
I have merged `master` into my branch. The only libclang bindings related change is now the introduction of `clang_getRange`.

